PR #3, opened 2021-01-08
CI: Add comment with App URL

anxolin:
Adds comment for PRaul: The links still won't work, cause they depend on the PRaul setup and S3 deployment

See the generated comment.


travis/deploy_pull_request.sh
```diff
@@ -39,9 +39,9 @@ function publish_pull_request_urls_in_github {
   else
     REVIEW_FEATURE_MESSAGE="\
 {\"body\": \"Travis automatic deployment:\\n \
-  * **🔭 [Explorer Dapp]($REVIEW_FEATURE_URL)**: Explorer test app\\n \
-  * **📈 [Trade Dapp]($REVIEW_FEATURE_URL/trade.html)**: Trade test app\\n \
-  * **📚 [Storybook](${REVIEW_FEATURE_URL}/storybook/)**: Component stories\""
+  * **🔭 [Explorer Dapp](${{ REVIEW_FEATURE_URL }})**: Explorer test app\\n \
+  * **📈 [Trade Dapp](${{ REVIEW_FEATURE_URL }})**: Trade test app\\n \
```
alfetopito, line 43: Intentional that both point to the same URL?

alfetopito (approved):
Approved if we want explorer and trade point to the same endpoint.

travis/deploy_pull_request.sh
anxolin: not at all... will fix

anxolin: mm, that's odd, I think u reviewed on old code or sth. it should look like this

image

alfetopito: o.O

alfetopito: Nope, that's the diff gh is giving me still
screenshot_2021-01-08_10-51-39

PR #8, opened 2021-01-11
web3: Update Provider to use request method

anxolin:
> Originally created in https://github.com/gnosis/gp-v1-ui/pull/1232
> Needs to be re-tested

Now that **web3** supports `request` method, this PR

- Adds **LogMiddleware** to debug `JsonRpcRequest|Response`, on by default in develop, disabled with `window.logMiddleware(false)`
- Enhances **composedProvider**, to handle`request` method if available on injected provider

Need to make it work with **WalletConnect** next

Please test with **Metamaks** (worked for me great) and dapp browsers

Part of #1199

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "react-toastify": "^5.5.0",
     "styled-components": "^5.0.0",
     "styled-theming": "^2.2.0",
-    "web3": "^1.2.10-rc.0",
+    "web3": "^1.2.11",
     "web3modal": "^1.6.3"
   },
   "devDependencies": {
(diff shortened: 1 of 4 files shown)


alfetopito:
So far so good.
Connected on xDai with Nifty wallet (Metamask fork) worked fine to place an order https://blockscout.com/poa/xdai/tx/0x43f83a9d0b1a498c893823d2819621cbe69f3a1b12d020c26de59c8c76078a69/internal-transactions

But this is branch seems outdated. I'll rebase with dev and try again.

PR #20, opened 2021-01-13
Theming - Add ThemedGlobalStyles

W3stside:
Wraps `GenericLayout` with `ThemedGlobalStyles` and our `ThemeProvider` boy which enables global theming. THere are several changes to make after this, including replacing all `var(...)` with our `theme.colourOfChoice` and others. Will do that on top of this

@alfetopito


src/components/layout/GenericLayout/index.tsx
```diff
@@ -10,9 +10,12 @@ export interface Props {
...
     <StaticGlobalStyle />
-    <Header menu={menu} tools={navTools} />
-    {children}
-    <Footer />
+    <ThemeProvider>
+      <ThemedGlobalStyle />
+      <Header menu={menu} tools={navTools} />
+      {children}
+      <Footer />
+    </ThemeProvider>
```
alfetopito, line 18: Just to be triple sure:

Any children will have the theme and global style already applied, **never again** `<ThemeProvider>` will be used, right?

alfetopito (approved):

src/components/layout/GenericLayout/index.tsx
W3stside: haha yes. you dont need to explicitly pass `ThemeProvider`, you're good.

Well, unless you want to. For example to change the entire theme before passing it in elsewhere. Or to pass a custom theme into it via `theme={customTheme}` prop.

src/components/layout/GenericLayout/index.tsx
```diff
@@ -10,9 +10,12 @@ export interface Props {
 export const GenericLayout: React.FC<Props> = ({ children, menu, navTools }) => (
   <div>
     <StaticGlobalStyle />
-    <Header menu={menu} tools={navTools} />
-    {children}
-    <Footer />
+    <ThemeProvider>
```
anxolin, line 13: Them and layout could be decoupled, and each app decides which ThemeProvider to use.
However, I think for now I think is good your current approach, we only have a single ThemeProvider and it's good not to head to ahead of ourselves

anxolin (approved):

src/components/layout/GenericLayout/index.tsx
W3stside: :+1:

PR #21, opened 2021-01-13
Theming/Fonts + Colours files more aware of responsibilties + minor style change

W3stside:
Merges into #20 

1. Moves `utils` for getting colours/fonts into those respective files
2. adds `font` sizes,faces,weight to theme to use (will be used to replace current `var(--font-weight-bold)` approach we are phasing out)
3. cleans up some styles


src/theme/styles/fonts.ts
```diff
@@ -24,5 +26,45 @@ const fontFace = css`
...
+}
+
+export function getFonts(mode: Theme): Fonts {
+  let FONT: Fonts
+
+  switch (mode) {
+    case Theme.LIGHT:
+    case Theme.DARK:
+      FONT = DEFAULT_FONT
+      break
```
alfetopito, line 62: Redundant, you might as well use only `default`, or skip the switch altogether and always return `DEFAULT_FONT`

alfetopito (approved):
Only tiny optimization suggestion

src/theme/styles/fonts.ts
W3stside: true, true, will fixito

src/theme/styles/global.ts
```diff
@@ -65,33 +58,26 @@ export const StaticGlobalStyle = createGlobalStyle`
...
-    text-decoration: underline;
-    cursor: pointer;
-      &:link, 
-      &:visited {
-        color: var(--color-text-active);
-      }
-  }
+  
 
+  /* TODO: move closer to H elements or sth */
```
anxolin, line 80: why? imo these styles that are not themed should be sth similar to a CSS reset imo:
https://meyerweb.com/eric/tools/css/reset/

anxolin (approved):
Great!

PR #22, opened 2021-01-13
Mock hooks on storybook

alfetopito:
# Summary

Adds hook mocking capabilities to storybook stories

# Usage

1. Create a file with the same name as the hook you want to mock on `src/storybook/mockedHooks/`
2. If you want your hook to be overridable on a per story/component basis, use `createMockHook` from `src/storybook/mockedHooks/mockHookContext.tsx`
3. Otherwise, simply implement a mocked version and export the same way as the original hook does

More details on `src/storybook/mockedHooks/mockHookContext.tsx` file

# :warning: Warning!

Based on https://github.com/gnosis/gp-ui/pull/20.
Merge that first to avoid confusion.


W3stside:
while this is cool and deffo overengineed dima style (compliment), there is a more simple fix of just wrapping our global state context around whatever you want to use. See `ThemeToggler` in `storybook/decorators`: https://github.com/gnosis/gp-ui/blob/0a6f506de25c0a4023a4d9a065e257e221b4fb58/src/storybook/decorators.tsx#L47

W3stside (approved):

alfetopito:
> while this is cool and deffo overengineed dima style (compliment), there is a more simple fix of just wrapping our global state context around whatever you want to use. See `ThemeToggler` in `storybook/decorators`:
> 
> https://github.com/gnosis/gp-ui/blob/0a6f506de25c0a4023a4d9a065e257e221b4fb58/src/storybook/decorators.tsx#L47

Of course there was a way of doing things that I missed :)

Makes sense will use that for having the theme working.

Still, I'd like to have the option to mock a hook when we want. It might do a network call or something else that we won't want to do on storybook.

I'll update the PR to NOT mock `useThemeManager`, but leave the setup ready.

.storybook/main.js
```diff
@@ -59,10 +56,28 @@ module.exports = {
...
+        plugins: [
+          ...config.resolve.plugins,
+          // Custom alias plugin required for webpack 4
+          // See https://github.com/webpack/webpack/issues/6817#issuecomment-542448438
+          new AliasPlugin(
+            'described-resolve',
+            [
+              {
+                name: 'hooks',
+                alias: [
```
anxolin, line 72: mm, I still didn't understand why u want to override the hook?
I mean, overriding a hook wouldn't be a red flag for a bad or complex module architecture?

anxolin (commented):
It looks great and it's nice we can mock hooks and all. My concern is in the direction of David's comment.

I see storybook as a nice way to test simple dumb components. These can have hooks, but initially I would think needing to mock it is a "smell" covering a bad design/architecture of components. I might be wrong, maybe u can mention some examples so I understand better.

Mocking a hook, breaks the "black box" principle. U need to know how a component is implemented, and if that changes, probably u need to update the mocked hook. 
Additionally, we modify the component. Any issue in the hook will be covered by that.

If it's for a "smart" component that uses a hook that gets data from redux, I would think:
- storybook for the dumb version could be enough
- otherwise, we can always just populate the data, to create the story scenarios. The data will arrive to the component.

.storybook/main.js
alfetopito: Think of it as a different approach to the same problem addressed by the API mocks we have.

Besides testing only dumb components, we could also use Storybook stories for implementing complex components composed of smaller dumber components, like full pages.

One example: think about the swap app's Swap page.
It relies heavily on hooks, aggregates several other small components and have some styled sneaking in some parts https://github.com/gnosis/gp-swap-ui/blob/develop/src/pages/Swap/index.tsx#L306

In order to make any UI changes, you'd normally start the change the code, refresh, etc.
But, depending on what step in the UI you want to test, there are network calls, intermediate states, and so on.
For this I think Storybook is of great value.
You can mock only key hooks (most of them can remain unchanged, depending on what they do behind the scenes) to get to the state you want and be able to view/modify only the UI.

I agree with the redux approach of providing the data, but there are also external calls we might not want to do on Storybook.

The reason why I wanted to bring it to this project no longer applies because there's another solution, and I don't have a use case for that right now.
More example I can give you are from CMM where smaller components had hooks, like this one:

Stories definition: https://github.com/gnosis/safe-cmm-app-react/blob/development/src/components/basic/display/MarketPrice/MarketPrice.stories.tsx
Actual stories: https://safe-cmm.gnosis.io/storybook/?path=/story/basic-display-marketprice--default
Main component: https://github.com/gnosis/safe-cmm-app-react/blob/development/src/components/basic/display/MarketPrice/index.tsx

The 'view' doesn't have any hooks and don't need any mocking.
The 'container' does, and there are also stories for it to "mock" the component behaviour under different hook responses.

Since this is sorta of a basic component - meaning, won't have a page by itself - it'll be included in another story, and there it'll need to mock it's inner hooks.

Having said all that, I can't say I'm happy with all design decisions we made in the CMM app and that there are no technical debts to address. 

Does that make it any clearer?

anxolin: Thanks for providing this detailed context, yes it makes it clearer. Just for the effort to document your thought I'm approving the PR ✅

By the way, storybook has also ways to intercept things like clicks, but I would also try to avoid that.
Okay, we'll see when we start to compose the components and we can decide in a per case basis if there's alternatives. 
Thanks!

anxolin (approved):

Issue #19, opened 2021-01-06
[v2] Consolidate UI redesign


anxolin:
This Epic is to keep track of the consolidation of the work on the classical UI. We want to extract and create from it reusable components we can use for a family of apps using similar look and feel. 

## Priorities
There's two priorities here:
* **High prio**: One is the for general layout and basic components, that will be used in our **Explorer**
* **Low Prio**: Everything else. Eventually is nice we can have the classical trading UI elements ready to use in GPv2, but probably Swap UI will be more important for the first months of the project.

## Task per component
Inside each prio, I break everything by component, so we can tackle one component at a time. Even one component can be tackled in multiple PR to make it easier for reviewers (address only one aspect of it).

In each reviewed component, we should make sure we check:
* Is the name correct? Is the path of the file ok? (see later namespaces)
* Is it dummy enough? Is the param API simple enough? Can we just inject some simple data and callbacks and use it?
* Does it have Storybook? Is the storybook good enough? Does the stories reflect the parametrization of the component?
* Does it have or need a smarter wrapper? Where does the data come from?
* Is the size/scope/complexity ok? does it need to be broken down
* Does it use the theme?
* Styles are ok? Also, if there's not A LOT of CSS on it, I would have the CSS in the component, not in a separated styled file.

## Namespaces
We have too many components. Some time ago, we started to group components in a directory of related components.
It would be good to have a clear picture of our namespaces, and make sure the components are in the right place.

For example, there's a tabular order book, a trades part of the order book, and a depth chart, maybe these 3 could be in the same namespace called "orderbook"

![image](https://user-images.githubusercontent.com/2352112/103770226-f1763500-5025-11eb-8f43-8f6852c448a8.png)

PR #30, opened 2021-01-14
Issue 19/generic layout

alfetopito:
# Summary

Part of #19 

Small refactor on `GenericLayout` component to make it more generic.

Taking header and footer as props instead of props of inner props.
Able even to disable header/footer if needed.

Also updated storybook

Is this in line with what was initially thought of?

# :warning:  Warning

Depends on #20
Merge after that one.


anxolin (approved):

src/apps/explorer/ExplorerApp.tsx
```diff
@@ -30,28 +31,32 @@ const Home = React.lazy(
     ),
 )
 
-export const SwapAppV1: React.FC = () => {
+const HEADER = (
+  <Header
+    menu={
```
W3stside, line 36: well Dima's idea and also mine was to avoid using slotted components: `menu` kind of thing. But tbh I'm not really decided yet on what approach i like best. In a way this is nice because it forces a certain layout (which contrarians would state is wrong) but meh I kinda like it

src/apps/explorer/ExplorerApp.tsx
```diff
@@ -30,28 +31,32 @@ const Home = React.lazy(
...
-        <Link to="/markets">Markets</Link>
-      </li>
-    </Menu>
-  )
-
   return (
     <>
       <Router basename={process.env.BASE_URL}>
-        <GenericLayout menu={menu}>
+        <GenericLayout header={HEADER}>
```
W3stside, line 59: why is HEADER all caps?

src/components/layout/GenericLayout/index.tsx
```diff
@@ -1,19 +1,41 @@
...
-import { StaticGlobalStyle } from 'theme'
-export interface Props {
-  menu?: React.ReactNode
-  navTools?: React.ReactNode
+import { Header } from './Header'
+
+export type Props = {
+  header?: JSX.Element | null
+  footer?: JSX.Element | null
+  children: React.ReactNode
```
W3stside, line 11: don't need this, FC passes children implicitly. could be a linter issue if you're seeing errors

src/components/layout/GenericLayout/index.tsx
```diff
@@ -1,19 +1,41 @@
...
+
+/**
+ * Generic layout with optional header and footer
+ * Applies global and theme styles to all children
+ *
+ * If not header/footer set, use default.
+ * If a custom passed in will be used instead.
+ * To remove header/footer, pass null
+ */
+export const GenericLayout: React.FC<Props> = (props) => {
```
W3stside, line 25:
```suggestion
export const GenericLayout: React.FC<Props> = ({
  header = defaultHeader,
  footer = defaultFooter,
  children,
}) => {
```

src/components/layout/GenericLayout/index.tsx
```diff
@@ -1,19 +1,41 @@
...
+  const { children, header, footer } = props
+
+  return (
+    <div>
+      <StaticGlobalStyle />
+      <ThemeProvider>
+        <ThemedGlobalStyle />
+        {header !== null && (header || defaultHeader)}
```
W3stside, line 33: if you accept the suggestion above, then:
```suggestion
        {header}
```

src/components/layout/GenericLayout/index.tsx
```diff
@@ -1,19 +1,41 @@
...
+        {children}
+        {footer !== null && (footer || defaultFooter)}
```
W3stside, line 35: if you accept my suggestions above, then:

```suggestion
        {footer}
```

W3stside (commented):
Nice, looks clean. minor comments

src/apps/explorer/ExplorerApp.tsx
alfetopito: Agreed. That's the idea for his refactor, just the `GenericLayout` component that I'm not using it.
If we want to make it even more generic, we can indeed have it only accept children, and have another component called `Layout` that does have header/footer.

I can go that way if preferred.

src/apps/explorer/ExplorerApp.tsx
alfetopito: To be explicit it's a static constant

src/components/layout/GenericLayout/index.tsx
alfetopito: Yeah, the storybook gets angry when I don't tell it has a `children` prop
I can use the `PropsWithChildren` though (I found out about this later)

src/components/layout/GenericLayout/index.tsx
alfetopito: The problem with this is that it'll overwrite `null`, won't it?

src/components/layout/GenericLayout/index.tsx
alfetopito: As said above, I still want to be able to disable the component on `null`, thus the condition.

src/components/layout/GenericLayout/index.tsx
alfetopito: I was wrong, it does work. I'll take it.

src/components/layout/GenericLayout/index.tsx
W3stside: hmmm okidoki

PR #31, opened 2021-01-14
Issue 19/generic layout header

alfetopito:
# Summary

Part of #19 

Waterfalls into https://github.com/gnosis/gp-ui/pull/30

Refactoring `GenericLayout/Header`.

Don't think this needs a storybook of its own.

# :warning: Warning

Don't merge before https://github.com/gnosis/gp-ui/pull/30

### DIFF
--- a/src/apps/explorer/ExplorerApp.tsx
+++ b/src/apps/explorer/ExplorerApp.tsx
@@ -32,21 +32,19 @@ const Home = React.lazy(
 )
 
 const HEADER = (
-  <Header
-    menu={
-      <Menu>
-        <li>
-          <Link to="/">Batches</Link>
-        </li>
-        <li>
-          <Link to="/trades">Trades</Link>
-        </li>
-        <li>
-          <Link to="/markets">Markets</Link>
-        </li>
-      </Menu>
-    }
-  />
+  <Header>
+    <Menu>
+      <li>
+        <Link to="/">Batches</Link>
+      </li>
+      <li>
+        <Link to="/trades">Trades</Link>
+      </li>
+      <li>
+        <Link to="/markets">Markets</Link>
+      </li>
+    </Menu>
+  </Header>
 )
 
 export const ExplorerApp: React.FC = () => {
--- a/src/components/layout/GenericLayout/Header/index.tsx
+++ b/src/components/layout/GenericLayout/Header/index.tsx
@@ -56,17 +56,11 @@ const Logo = styled(Link)`
   }
 `
 
-interface Props {
-  menu?: React.ReactNode
-  tools?: React.ReactNode
-}
-
-export const Header: React.FC<Props> = ({ menu, tools }) => (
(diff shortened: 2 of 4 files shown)


src/apps/explorer/ExplorerApp.tsx
```diff
@@ -32,21 +32,19 @@ const Home = React.lazy(
...
-        <li>
-          <Link to="/trades">Trades</Link>
-        </li>
-        <li>
-          <Link to="/markets">Markets</Link>
-        </li>
-      </Menu>
-    }
-  />
+  <Header>
```
W3stside, line 35: ooohhhhh i see now. ignore previous comment on slotted things :sweat_smile:

W3stside (approved):

src/components/layout/GenericLayout/Header/index.tsx
```diff
@@ -56,17 +56,11 @@ const Logo = styled(Link)`
...
 
-interface Props {
-  menu?: React.ReactNode
-  tools?: React.ReactNode
-}
-
-export const Header: React.FC<Props> = ({ menu, tools }) => (
+export const Header: React.FC = ({ children }) => (
   <HeaderStyled>
     <Logo to="/" href="#">
```
anxolin, line 61: Shouldn't be the logo out too?

src/apps/explorer/ExplorerApp.tsx
alfetopito: Already replied :D
Will make sure to ignore your reply if you ignore mine :)

src/components/layout/GenericLayout/Header/index.tsx
alfetopito: I thought the logo was part of the generic header, no?

W3stside: I have some concerns regarding this... at what point is the reusablity/generic-icity(?) of the component(s) too much? Seeing as this is indeed `gp-ui`, emphasis on `gp` shouldn't some things be opinionated? In this case I'm referring to the `Logo` and maybe the placement of the wallet/menu stuff as it's always the same `Header` passed around `gp` apps.

Thoughts?

alfetopito: I second that, and it's the point I raised last week in the chat https://gnosisinc.slack.com/archives/G01J17B9G6N/p1610737288016700

anxolin: My point is that this component is mixing too approaches. We had some discussion about this, I I think the conclusion was that for this container layout components, we try not to mix approaches. We either provide the content as children, or as parametrization of components.

I mean, it feels natural seeing the code you have to just pass the logo with the children, and it's more flexible. We can make small logo variations...

Anyways, I don't care that much about this, if we change one way or the other won't matter much. We can re-work later

anxolin (approved):

PR #34, opened 2021-01-15
Test ci split

anxolin:
Just doing some experimentation in case we can reduce the deployment times:

image

It was not a great performance boost, but improved a bit and it will be more detailed. Now when we fail, we can see more clear what job failed
image

image

## Nice additional thing in this PR
Now the WEB is downloadable! so we can get the ZIP and run it in any static server:
image


alfetopito (approved):
Looks great.

But does it mean that every step does a checkout and `yarn install` independently?
Is it possible to somehow re-use those steps to avoid duplication?

anxolin:
> But does it mean that every step does a checkout and yarn install independently?

checkout almost all jobs yes, but that's fast
install only the setup. The others use the cache...
I see the overhead, but overall seemed faster

> Is it possible to somehow re-use those steps to avoid duplication?

The only way I found is, with cache for the node_modules, and for the websites I use artifacts

W3stside (approved):
nice job!

PR #70, opened 2021-01-18
Auto-Merge setup via Mergify

W3stside:
Using mergify setup from dex-services

Merges when actions are green, base branch is ~`master`~ `develop`

~@anxolin base branch bring `master` is ok?~


alfetopito (commented):
I think we should be based on `dev` since that's where we work and will be easier to see if something breaks that's not caught by the tests.

.mergify.yaml
```diff
@@ -0,0 +1,23 @@
+pull_request_rules:
+  - name: Merge approved and green PRs with merge-when-green label
+    conditions:
+      - "#approved-reviews-by>=1"
+      - status-success=Travis CI - Pull Request
```
anxolin, line 5: Why travis? we us GH now

.mergify.yaml
```diff
@@ -0,0 +1,23 @@
...
+      - label=merge when green
+    actions:
+      merge:
+        method: squash
+        strict: smart+fasttrack
+        commit_message: title+body
+  - name: automatic merge for Dependabot pull requests
+    conditions:
+      - author~=^dependabot(|-preview)\[bot\]$
+      - status-success=Travis CI - Pull Request
```
anxolin, line 16: no travis

.mergify.yaml
```diff
@@ -0,0 +1,23 @@
...
+      - base=develop
+    actions:
+      merge:
+        method: squash
+        strict: smart+fasttrack
+        commit_message: title+body
```
anxolin, line 22: Does it add a break line between the title and the body?

anxolin (commented):
Awesome! thanks @W3stside

anxolin:
Once it works, can u set it up in https://github.com/gnosis/gp-swap-ui too pls?

.mergify.yaml
W3stside: yup, will fix. Actually this whole section i am removing as it's ALL PRs, and I just want Dependabot

.mergify.yaml
W3stside: :+1:

.mergify.yaml
W3stside: need to check

alfetopito (approved):

.mergify.yaml
W3stside: i'll check against one of the open dependabot PRs

anxolin (approved):
Great!

Issue #40, opened 2021-01-18
[v2] Generic Table


anxolin:
Take a look at "ActiveOrdersContent" and review if we can refactor the styles for a generic table we can use to display results of the batch explorer

PR #71, opened 2021-01-19
Simple Table

anxolin:
Part of #40 

Proposal of a simple component for simple tables

We don't need to get to ahead of ourselves, my suggestion is just reuse the styles of @biocom and allow to inject the data

The component receives a header, and a the rows.

U can see how I moved some styles out of the generic component, mainly the LONG/SHORT colors, because this belong to the specific component inheriting the table.

image


biocom:
Just as a side note: The goal here was to keep a semantic HTML <table> but to enhance it with CSS Grid for layout purposes. Mainly to have greater control on how cells are positioned.

src/components/common/SimpleTable/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+    &:first-of-type {
+      padding-left: 1rem;
+      justify-content: flex-start;
+    }
+  }
+
+  > thead > tr,
+  > tbody > tr {
+    align-items: center;
+    ${({ $numColumns }): string => ($numColumns ? `grid-template-columns: repeat(${$numColumns}, 1fr)` : '')}
```
anxolin, line 95: Maybe the `;` is missing?
```suggestion
    ${({ $numColumns }): string => ($numColumns ? `grid-template-columns: repeat(${$numColumns}, 1fr);` : '')}
```

anxolin (commented):
Approved 👌 
Thanks Leandro!

src/components/common/SimpleTable/index.tsx
W3stside:
```suggestion
    ${({ $numColumns }): string | false => $numColumns && `grid-template-columns: repeat(${$numColumns}, 1fr);`}
```

src/components/common/SimpleTable/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+export type Props = PropsWithChildren<{
+  header: JSX.Element
+  className?: string
+  numColumns?: number
+}>
+
+export const SimpleTable: React.FC<Props> = ({ header, children, className, numColumns }) => (
+  <Wrapper $numColumns={numColumns} className={className}>
+    <thead>{header}</thead>
+    <tbody>{children}</tbody>
```
W3stside, line 108: if we're using a prop for the table header, the children should be renamed to `body` or better even `tableHeader` and `tableBody` imo

src/components/common/SimpleTable/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+  padding: 0;
+  box-sizing: border-box;
+  width: 100%;
+  border-spacing: 0;
+  display: inline-grid;
+  grid-template-areas:
+    'head-fixed'
+    'body-scrollable';
+
+  tr {
```
W3stside, line 19: please change these to nested props:
```scss
tr {
  text-align: left;
  padding: 0;

  > td {
    padding: 0;
    ....
  }
```

makes it much easier to reason with and edit later

src/components/common/SimpleTable/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+    text-align: left;
+    padding: 0;
+  }
+
+  tr > td {
```
W3stside, line 24: same thing here, nesting css properties

src/components/common/SimpleTable/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+  > thead {
+    grid-area: head-fixed;
+    position: sticky;
+    top: 0;
+    height: auto;
+    display: flex;
+    align-items: center;
+  }
+
+  > thead > tr {
```
W3stside, line 40: nesting please. won't comment anymore on these i think you get it

src/components/common/SimpleTable/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  > thead > tr > th {
+    font-weight: var(--font-weight-normal);
+    &:not(:first-of-type) {
+      text-align: right;
+    }
+  }
+
+  > tbody {
+    grid-area: body-scrollable;
```
W3stside, line 54: cool

W3stside (changes requested):

src/components/common/SimpleTable/index.tsx
alfetopito: I thought about that because it bothers me too.
But in the end I chose to let it as is to allow a more "natural" <Table> experience, except for the header.

@anxolin your child, what do you think? Don't care? 

I'm fine changing if there's a strong position here.

src/components/common/SimpleTable/index.tsx
alfetopito: Actually, I'll revert David's suggestion because of types and just use Anxo's `;` fix.
screenshot_2021-01-20_10-10-38

alfetopito (approved):
Self approve, because this was started by Anxo

biocom:
Tested the component in the StoryBook. Think it's good. The only thing I could find was that in the `SimpleTable__Wrapper` we have this style (which I added originally):
`height: calc(100% - 8.8rem);`

This was basically to make it work in context of the classic trading interface. Think we should change this to `height: 100%` in case we want to make it a generic table. 

One way to properly test the inner scroll feature, is to set the parent DIV (where the SimpleTable is child of) to a height of say `300px`. When I tested it didn't utilize the full height of the parent DIV because of the above CSS rule.

W3stside:
> Tested the component in the StoryBook. Think it's good. The only thing I could find was that in the `SimpleTable__Wrapper` we have this style (which I added originally):
> `height: calc(100% - 8.8rem);`
> 
> This was basically to make it work in context of the classic trading interface. Think we should change this to `height: 100%` in case we want to make it a generic table.
> 
> One way to properly test the inner scroll feature, is to set the parent DIV (where the SimpleTable is child of) to a height of say `300px`. When I tested it didn't utilize the full height of the parent DIV because of the above CSS rule.

seconding this

W3stside (changes requested):
Changes to table height as @biocom made

src/components/common/SimpleTable/index.tsx
W3stside: my change my change my change :rocket:

alfetopito: ok ok, I'll do it

W3stside (approved):

Issue #41, opened 2021-01-18
[v2] Generic tabs


anxolin:
* [ ] Review the tabs components and make sure we have a simple dumb tabs component
![image](https://user-images.githubusercontent.com/2352112/103770274-04890500-5026-11eb-824e-7229cd42ee73.png)
* [ ] src/components/common/Tabs

biocom:
Just as a note (but you may already be aware): Tabs component also in use for [Orderbook / Trades] and [Price Chart / Depth Chart]

PR #72, opened 2021-01-19
[Generic Tabs: Part 1] - Cleanup (logic, types)

W3stside:
Part of #41 

This is the first part, I am cleaning up the current implementation by Michel which is already very good. Best to do this now before PR gets bigger

Happy to take comments here but i dont see `common/Tabs` - the generic tab part being too different from the already implements `TabItem` file

### DIFF
--- a/src/components/OrdersWidgetDemo/index.tsx
+++ b/src/components/OrdersWidgetDemo/index.tsx
@@ -23,12 +23,13 @@ const tabItems: TabItemInterface[] = [
 
 // Provide a custom theme
 const tabThemeConfig = getTabTheme({
-  activeBg: '--color-primary',
-  activeText: '--color-text-primary',
-  inactiveBg: '--color-transparent',
-  inactiveText: '--color-text-secondary2',
-  fontWeight: '--font-weight-normal',
-  fontSize: '--font-size-default',
+  activeBg: 'var(--color-primary)',
+  activeText: 'var(--color-text-primary)',
+  inactiveBg: 'var(--color-transparent)',
+  inactiveText: 'var(--color-text-secondary2)',
+  activeBorder: 'none',
+  fontWeight: 'var(--font-weight-normal)',
+  fontSize: 'var(--font-size-default)',
   letterSpacing: '0.03rem',
 })
 
--- a/src/components/OrdersWidgetDemo/index.tsx
+++ b/src/components/OrdersWidgetDemo/index.tsx
@@ -1,26 +1,23 @@
 import React from 'react'
-import Tabs, { getTabTheme, TabItemType } from 'components/common/Tabs/Tabs'
+import Tabs, { getTabTheme, TabItemInterface } from 'components/common/Tabs/Tabs'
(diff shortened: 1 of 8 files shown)


alfetopito (approved):
Storybook will come next, yes?

src/components/OrdersWidgetDemo/index.tsx
```diff
@@ -1,39 +1,37 @@
...
+import Tabs, { getTabTheme, TabItemInterface } from 'components/common/Tabs/Tabs'
 import { OrdersWidgetDemo as Wrapper } from './OrdersWidgetDemo.styled'
 import { ActiveOrdersContent } from './ActiveOrdersContent'
 
-const tabItems: TabItemType[] = [
+const tabItems: TabItemInterface[] = [
   {
     id: 1,
-    title: 'Active Orders',
+    tab: 'Active Orders: ' + 5,
```
anxolin, line 9: nit: why tab is a better name than title? I like it more with title. Agree on removing `count`.

src/components/common/Tabs/Tabs.tsx
```diff
@@ -5,28 +5,31 @@ import styled from 'styled-components'
 import TabItem from 'components/common/Tabs/TabItem'
 import TabContent from 'components/common/Tabs/TabContent'
 
-export interface TabItemType {
-  readonly id: number
-  readonly title: string
+type TabId = number
+
+export interface TabItemInterface {
```
anxolin, line 10: Why a type need a sufix `Type` or `Interface` at the end? is `TabItem` not good?

anxolin (approved):
cool

src/components/OrdersWidgetDemo/index.tsx
```diff
@@ -1,39 +1,37 @@
...
     content: <ActiveOrdersContent />,
-    count: 5,
   },
   {
     id: 2,
-    title: 'Order History',
+    tab: 'Order History: ' + 10,
```
biocom, line 14: What is the reason for adding a number here?

src/components/OrdersWidgetDemo/index.tsx
W3stside: because before `title: string` now `tab: ReactNode` which is a better vernacular imo

src/components/common/Tabs/Tabs.tsx
W3stside: `TabItem` as a component already exists, was getitng name clash

src/components/OrdersWidgetDemo/index.tsx
W3stside: you had `count` before which wasn't used, as well as opinionated. This is to show count and not need a prop inside the `Tab` generic which shouldn't be aware of it. Leave it to the passed in `tab` prop to deal with that

biocom: Ah got it 👍🏼

biocom (approved):

PR #75, opened 2021-01-19
Table stories

anxolin:
WIP

The idea is to show how to create tables with different number of columns.

We could also create a story where we don't provide the number of columns, but instead define using CSS the columns (extending the styled component)


alfetopito:
I'll merge this onto #71 and finalize the task there

PR #78, opened 2021-01-20
Smart Table [Resizable]

W3stside:
The main thing here to showcase is the resizability - the actual <table> layout/format etc can be changed. Plan is to adapt it to @alfetopito already made stuff

I want to see what you guys think about this!

You can test it in a separate route for now, didn't make a story for it:
`http://localhost:8080/smart`

### Try resizing the headers (you'll see a span pop in on hover)

### DIFF
--- a/src/components/common/SmartTable/index.tsx
+++ b/src/components/common/SmartTable/index.tsx
@@ -1,0 +1,181 @@
+import React, { RefObject } from 'react'
+import styled from 'styled-components'
+import { useSmartTable } from './hooks'
+
+const ResizableSpan = styled.span`
+  position: absolute;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  background: black;
+  opacity: 0;
+  width: 5%;
+  cursor: col-resize;
+
+  &:hover {
+    opacity: 0.5;
+  }
+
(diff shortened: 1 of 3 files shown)


W3stside:
**TOP PRIORITY!!!** :bomb: :bomb: :bomb: :bomb:

src/components/common/SmartTable/index.tsx
```diff
@@ -0,0 +1,181 @@
...
+    height: 12rem;
+  }
+
+  /* TODO: RECONSIDER - TESTING */
+  tr:nth-child(even) td {
+    background: #f8f6ff;
+  }
+`
+
+export const SmartTable: React.FC<{ headerChildren: React.ReactNode[]; className?: string }> = (props) => {
```
anxolin, line 95: We usually define and export an interface called `Prop`, handy for the storybook

anxolin (commented):
Not sure we need it right now, but it was fun to try out.

It looks very similar to, this is where u got the code for the hook?:
https://webdevtrick.com/resizable-table-columns

Probably if we ever need fancy features we also have the option of libraries like https://react-table.tanstack.com
https://react-table.tanstack.com

anxolin:
@W3stside Dave, can we close?

PR #80, opened 2021-01-21
[GP-V1::FIX] - use old MEDIA queries

W3stside:
Fixes broken media queries in gp-v1 due to using `applyMediaStyles` which requires `theme`


anxolin (approved):
Works, thanks

alfetopito (approved):
Alright then, re-using those components is off the table for the new apps.

PR #81, opened 2021-01-21
Load apis per app

alfetopito:
# Summary

Split API loading per APP.

Right now there are only `gp-v1` APIs, so all apps are empty.

# Notes

Not very happy with this either.
I would like to conditionally load APIs based on the `APP` env var.
But that might cause the APIs to be potentialy `undefined`, which will make the refactor a nightmare.

Any better ideas?


W3stside:
Screenshot from 2021-01-21 19-26-30
:scream: :skull:  :skull: RIP

src/api/wallet/composeProvider.ts
```diff
@@ -10,7 +10,7 @@ import { TransactionConfig } from 'web3-core'
 import { numberToHex, hexToNumber } from 'web3-utils'
 import { isWalletConnectProvider, Provider } from './providerUtils'
 import { logDebug } from 'utils'
-import { web3 } from 'api'
+import { web3 } from 'api/gpV1'
```
W3stside, line 13: would be cool to have one central web3 api

src/apps/gp-v1/pages/Settings.tsx
```diff
@@ -3,7 +3,7 @@ import { useForm, Resolver as FormResolver, FieldErrors, ResolverResult } from '
 import styled from 'styled-components'
 import { DevTool } from 'HookFormDevtool'
 
-import { walletApi } from 'api'
+import { walletApi } from 'api/gpV1'
```
W3stside, line 6: and central walletapi also would be kewl

W3stside:
i dont hate it

W3stside:
but

W3stside:
i dont love it neither

src/api/wallet/composeProvider.ts
alfetopito: Well, every app can export their own web instance exactly the same way as `gp-v1` does.

But I get your point. Every app will need `web3`. Will change that.

src/apps/gp-v1/pages/Settings.tsx
alfetopito: Same point as for web3, except, will for instance the explorer need a wallet api?

alfetopito:
> i dont love it neither

Same.
Any other suggestions?

W3stside:
Apart from creating a gp-v1-api repo and importing those there, nope. I was
trolling low key before, this is good

src/PrivateRoute.tsx
```diff
@@ -12,7 +12,7 @@ const useIsWalletConnected = (): {
   useEffect(() => {
     import(
       /* webpackChunkName: "API_chunk"*/
-      'api'
+      'apps/gp-v1/api'
     )
       .then(({ walletApi }) => walletApi.isConnected())
       .then((isConnected) => {
```
alfetopito, line 18: Next refactor step would be to split the hooks between shared/common and app specific.

Then we can import common from base /api instances and so on.

src/api/index.ts
```diff
@@ -181,17 +181,5 @@ function createTcrApi(web3: Web3): TcrApi | undefined {
   return tcrApi
 }
 
-// Build APIs
```
anxolin, line 184: good call on not building all the APIs

src/apps/gp-v1/api.ts
```diff
@@ -0,0 +1,39 @@
+import {
```
anxolin, line 1: I like that you moved this to the app! good call

anxolin (approved):
The PR looks good to me! I think u did the right call moving things to their folder.

But for the new APPs, altough I see we can continue the pattern of using folders in api/<integration> , I think we can explore the approach of using plain old functions instead of the Class approach we took last time. 

Like https://github.com/gnosis/gp-swap-ui/blob/develop/src/custom/utils/operator.ts
Maybe I can even make a PR with the suggestion.

Using a functional approach is nicer for tree-shaking, specially the APIs have a lot of method, and I think at some points it got a bit too much (too big classes with hierarchies, etc).

By the way, we can bring this file

WDYT?

anxolin:
> Not very happy with this either.
It's fine :) 

>I would like to conditionally load APIs based on the APP env var. But that might cause the APIs to be potentially undefined, which will make the refactor a nightmare.

I don't think we should do this. Every app pulls the APIs he wants. Actually, I'm suggesting pooling the methods directly instead of using classes. this way each component/hook decides:

```ts
import { crazyMethod } from 'api/madService'
```

W3stside:
Interestingly, the functional API approach is how Dima and I coded the
first projects. (We both read YDKJS haha)

anxolin:
> Interestingly, the functional API approach is how Dima and I coded the
first projects. (We both read YDKJS haha)

This means you also prefer to do new APIs as functions? we were coding most of things functionally, the APIs were done like this just because the idea was to have mock/impl and respecting the same interface, and be able to switch one for the other. 
I still think this is useful, but some API got bigger and more complex and it feels strange that's the only place we use OO.

W3stside:
> > Interestingly, the functional API approach is how Dima and I coded the
> > first projects. (We both read YDKJS haha)
> 
> This means you also prefer to do new APIs as functions? we were coding most of things functionally, the APIs were done like this just because the idea was to have mock/impl and respecting the same interface, and be able to switch one for the other.
> I still think this is useful, but some API got bigger and more complex and it feels strange that's the only place we use OO.

Yes I think the classless approach using singletons is the way forward

W3stside (approved):

PR #82, opened 2021-01-22
Apì operator

anxolin:
image

Brings the operator API from the swap UI.
It's just a proposal on how to do new APIs. I'm not super convince we should continue for the new APIs using classes. They become too big and less tree-shakable 

This has to do with my comment here https://github.com/gnosis/gp-ui/pull/81#pullrequestreview-574084789


alfetopito (approved):
I like it, let's go for this API approach.

alfetopito:
Actually, now that I'm implementing it it came to me.
How to mock stuff without explicitly importing the mock functions?

Sorta like we do with API instances to 'inject' the version we want when we want?
Guess we'll have to register the functions in a central place like we do right now with the API class instances... :thinking:

W3stside:
> Actually, now that I'm implementing it it came to me.
> How to mock stuff without explicitly importing the mock functions?
> 
> Sorta like we do with API instances to 'inject' the version we want when we want?
> Guess we'll have to register the functions in a central place like we do right now with the API class instances...

With normal functions as our API, we will likely adopt a singleton approach, with an initiation function, something like:
```ts
let API: AppApi | Falsy
export const appApi = async ({ provider, forceRefresh, useMock }: { provider?: Provider, forceRefresh?: boolean, useMock?: boolean }) => {
  if (useMock) return mockApi
  if (API && !forceRefresh) return API

  API = await initAPI(provider, forceRefresh)
  return API
}
```
Wdyt?

alfetopito:
> > Actually, now that I'm implementing it it came to me.
> > How to mock stuff without explicitly importing the mock functions?
> > Sorta like we do with API instances to 'inject' the version we want when we want?
> > Guess we'll have to register the functions in a central place like we do right now with the API class instances...
> 
> With normal functions as our API, we will likely adopt a singleton approach, with an initiation function, something like:
> 
> ```ts
> let API: AppApi | Falsy
> export const appApi = async ({ provider, forceRefresh, useMock }: { provider?: Provider, forceRefresh?: boolean, useMock?: boolean }) => {
>   if (useMock) return mockApi
>   if (API && !forceRefresh) return API
> 
>   API = await initAPI(provider, forceRefresh)
>   return API
> }
> ```
> 
> Wdyt?

That might be a better approach than what I was trying to do.

Trying to replace original functions with mocked versions based on env var, similar to what we have on our class based apis. Not having much success:
Based on this PR https://github.com/gnosis/gp-ui/pull/82#issuecomment-765714367
The following change is to this file https://github.com/gnosis/gp-ui/pull/82/files#diff-eb2d8ad3a8627e5ba472e6de8e97ea7a12d51b86169d706f5d417aa862f56023R1

```typescript
import * as realApi from './operatorApi'
import * as mockApi from './operatorMock'
// realApi is read only. Making a copy
const toExport = { ...realApi }
if (process.env.MOCK_OPERATOR) {
  // Replacing real implementations with mocked versions
  Object.keys(mockApi).forEach((mockExport) => (toExport[mockExport] = mockApi[mockExport]))
}
// Re-exporting the result, mocked or not
export const { ...all } = toExport
```
---

Back to your idea, would every called need to "init" the function or do we intend to have a central place to do that like with class based apis?

PR #83, opened 2021-01-22
62/order detail page

alfetopito:
# Summary

Part of #62 

Only the page skeleton and the route
screenshot_2021-01-22_10-23-22

# Testing
Check it out on https://pr83--gpui.review.gnosisdev.com/order/asdajksh/

PR #89, opened 2021-01-25
62/api

### LINKED ISSUE #63: [view meta tx] Load data from API

alfetopito:
# Summary

Based on https://github.com/gnosis/gp-ui/pull/82 and #83 

Closes #63 

- Adds `getOrder` and `getOrders` apis, with mocks.
- Added some :star2: magic :star2: to export functions or mocks according to env var

screenshot_2021-01-25_12-55-46

Still looks awful. Will work on that next.

### DIFF
--- a/src/api/operator/index.ts
+++ b/src/api/operator/index.ts
@@ -2,7 +2,7 @@ import * as realApi from './operatorApi'
 import * as mockApi from './operatorMock'
 export * from './types'
 
-const useMock = process.env.MOCK_OPERATOR
+const useMock = process.env.MOCK_OPERATOR === 'true'
 
 // Re-exporting the result, mocked or not.
 // Unfortunately, did not find a way to export
--- a/src/api/operator/index.ts
+++ b/src/api/operator/index.ts
@@ -16,6 +16,5 @@ export const {
   getOrderLink = realApi.getOrderLink,
   postSignedOrder = realApi.postSignedOrder,
   getFeeQuote = realApi.getFeeQuote,
-  // function that only have a mock
-  getTrades = mockApi.getTrades,
+  // functions that only have a mock
 } = useMock ? { ...mockApi } : { ...realApi }
--- a/src/api/operator/index.ts
+++ b/src/api/operator/index.ts
@@ -1,2 +1,21 @@
-export * from './operatorApi'
-export * from './operatorMock'
+import * as realApi from './operatorApi'
+import * as mockApi from './operatorMock'
+export * from './types'
+
+const useMock = process.env.MOCK_OPERATOR
+
+// Re-exporting the result, mocked or not.
+// Unfortunately, did not find a way to export
+// a mix of mock/real implementations
+// without manually naming the exports
+export const {
+  // functions that have a mock
+  getOrder,
+  getOrders,
+  // functions that do not have a mock
+  getOrderLink = realApi.getOrderLink,
+  postSignedOrder = realApi.postSignedOrder,
+  getFeeQuote = realApi.getFeeQuote,
+  // function that only have a mock
+  getTrades = mockApi.getTrades,
+} = useMock ? { ...mockApi } : { ...realApi }
--- a/src/api/operator/operatorApi.ts
+++ b/src/api/operator/operatorApi.ts
@@ -33,22 +33,6 @@ function _getApiBaseUrl(networkId: Network): string {
   }
 }
 
-/**
- * Given a list of (key,value), return a URL query string prepended by `?`
- * Filters out values that are undefined. Empty string is accepted
- *
- * @param params List of key value pairs
- */
-function buildQueryString(params: [key: string, value?: string][]): string {
-  return (
-    '?' +
-    params
-      .filter(([, value]) => value !== undefined)
-      .map(([key, value]) => `${key}=${value}`)
-      .join('&')
-  )
-}
-
 export function getOrderLink(networkId: Network, orderId: OrderID): string {
   const baseUrl = _getApiBaseUrl(networkId)
 
@@ -135,7 +119,7 @@ export async function postSignedOrder(params: { networkId: Network; order: Order
   // Call API
   const response = await _post(networkId, `/orders`, order)
 
-  // Handle respose
+  // Handle response
   if (!response.ok) {
     // Raise an exception
     const errorMessage = await _getErrorForUnsuccessfulPostOrder(response)
@@ -198,19 +182,24 @@ export async function getOrder(params: GetOrderParams): Promise<RawOrder> {
  *  - buyToken: address
  */
 export async function getOrders(params: GetOrdersParams): Promise<RawOrder[]> {
-  const { networkId, owner, sellToken, buyToken } = params
+  const { networkId, ...searchParams } = params
+  const { owner, sellToken, buyToken } = searchParams
 
   console.log(
     `[getOrders] Fetching orders on network ${networkId} with filters: owner=${owner} sellToken=${sellToken} buyToken=${buyToken}`,
   )
 
-  const queryString =
-    '/orders/' +
-    buildQueryString([
-      ['owner', owner],
-      ['sellToken', sellToken],
-      ['buyToken', buyToken],
-    ])
+  const searchString = new URLSearchParams(
+    Object.keys(searchParams).reduce((acc, key) => {
+      // Pick keys that have values non-falsy
+      if (searchParams[key]) {
+        acc[key] = encodeURIComponent(searchParams[key])
+      }
+      return acc
+    }, {}),
+  ).toString()
+
+  const queryString = '/orders/?' + searchString
 
   const response = await _get(networkId, queryString)
 
--- a/src/api/operator/operatorApi.ts
+++ b/src/api/operator/operatorApi.ts
@@ -33,6 +33,41 @@ export interface FeeInformation {
   feeRatio: number
 }
 
+export type OrderKind = 'sell' | 'buy'
+
+// Raw API response
+export type RawOrder = {
+  creationDate: string
+  owner: string
+  uid: string
+  executedBuyAmount: string
+  executedSellAmount: string
+  executedFeeAmount: string
+  invalidated: boolean
+  sellToken: string
+  buyToken: string
+  sellAmount: string
+  buyAmount: string
+  validTo: number
+  appData: number
+  feeAmount: string
+  kind: OrderKind
+  partiallyFillable: boolean
+  signature: string
+}
+
+type WithNetworkId = { networkId: Network }
+
+export type GetOrderParams = WithNetworkId & {
+  orderId: string
+}
+
+export type GetOrdersParams = WithNetworkId & {
+  owner?: string
+  sellToken?: string
+  buyToken?: string
+}
+
 function _getApiBaseUrl(networkId: Network): string {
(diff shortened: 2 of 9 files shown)


W3stside:
15 file changes? Is the base branch correct?

src/api/operator/index.ts
```diff
@@ -0,0 +1,21 @@
...
+  // functions that have a mock
+  getOrder,
+  getOrders,
+  // functions that do not have a mock
+  getOrderLink = realApi.getOrderLink,
+  postSignedOrder = realApi.postSignedOrder,
+  getFeeQuote = realApi.getFeeQuote,
+  // function that only have a mock
+  getTrades = mockApi.getTrades,
+} = useMock ? { ...mockApi } : { ...realApi }
```
anxolin, line 21: Aha, ok, u want to maintain the useMock thing. 

My initial idea was more simple, just mock the services that are not yet implemented. But happy to do it this way too if we think is useful

src/apps/explorer/components/OrderWidget/index.tsx
```diff
@@ -0,0 +1,30 @@
+import React, { useEffect, useState } from 'react'
```
anxolin, line 1: Cool, I like we use the widget notation for smart components

At a later PR I would love to see a dumb dumb one receiving just the order data

src/apps/explorer/pages/Home.tsx
```diff
@@ -11,11 +13,30 @@ const Wrapper = styled.div`
...
-    <h1>Gnosis Protocol Explorer</h1>
-    <p>Welcome :)</p>
-  </Wrapper>
-)
+export const Home: React.FC = () => {
+  const [fee, setFee] = useState('')
+  const [trades, setTrades] = useState<string[]>([])
+
+  useEffect(() => {
+    getFeeQuote(Network.Mainnet, '0x6810e776880c02933d47db1b9fc05908e5386b96')
```
anxolin, line 21: U can kill this if u want. It was just to play around

anxolin (approved):
Great!

src/apps/explorer/components/OrderWidget/index.tsx
alfetopito: Yes, I left a TODO to split it later.
Mainly when we start to really wire it up. 
Right now is mostly a placeholder.

src/apps/explorer/pages/Home.tsx
alfetopito: Ok, will remove.

alfetopito:
> 15 file changes? Is the base branch correct?

As I mentioned in the description ;)

> Based on #82 and #83

Thus, the noise.
Once those are approved/merged I'll rebase to make it look pretty.

src/api/operator/index.ts
alfetopito: Yeah, we might need to mock stuff that's not yet available or hard to reproduce, such as the `getOrder` calls that disappear very quickly.


Not very happy with this solution though, as every new function will need to be manually added to this list of exports.
Guess that's a down side of the functional style api?

anxolin:
@W3stside this one could use one review of yours

src/api/operator/index.ts
W3stside: why spread into object instead of returning the object directly?

src/api/operator/operatorApi.ts
```diff
@@ -43,6 +33,22 @@ function _getApiBaseUrl(networkId: Network): string {
   }
 }
 
+/**
+ * Given a list of (key,value), return a URL query string prepended by `?`
+ * Filters out values that are undefined. Empty string is accepted
+ *
+ * @param params List of key value pairs
+ */
+function buildQueryString(params: [key: string, value?: string][]): string {
```
W3stside, line 42: is this sth the `URL` or `URLSearchParams` apis can do? if not then ignore me

src/api/operator/operatorApi.ts
```diff
@@ -162,3 +168,57 @@ export async function getFeeQuote(networkId: Network, tokenAddress: string): Pro
...
+}
+
+/**
+ * Gets a list of orders
+ * Optional filters:
+ *  - owner: address
+ *  - sellToken: address
+ *  - buyToken: address
+ */
+export async function getOrders(params: GetOrdersParams): Promise<RawOrder[]> {
```
W3stside, line 200: how are you handling these `async` methods if it throws for any reason not relayed to `!response.ok` - could this ever happen? I would suggest `try/catch` if so

src/apps/explorer/components/OrderWidget/index.tsx
```diff
@@ -1,9 +1,30 @@
...
+  const [error, setError] = useState('')
+
   const { orderId } = useParams<{ orderId: string }>()
 
-  return <div>You are viewing order id: {orderId}</div>
+  useEffect(() => {
+    // TODO: get network from a hook or something
+    getOrder({ networkId: 4, orderId })
+      .then((order) => {
+        console.log('got order', order)
```
W3stside, line 16: just for testing right? maybe a TODO to remove in case u forget

src/apps/explorer/components/OrderWidget/index.tsx
```diff
@@ -1,9 +1,30 @@
...
+        setError('')
+      })
+      .catch((e) => setError(e.message))
+  }, [orderId])
+
+  return (
+    <div>
+      <h1>Details for order id: {orderId}</h1>
+      {order && <p>{JSON.stringify(order)}</p>}
+      {error && <p>{error}</p>}
```
W3stside, line 27: I would consider making the error closeable

W3stside: def not for this PR just thinking out loud

W3stside: so maybe add a `// TODO ...` ? :thinking:

src/api/operator/operatorApi.ts
```diff
@@ -162,3 +168,57 @@ export async function getFeeQuote(networkId: Network, tokenAddress: string): Pro
...
+
+  const response = await _get(networkId, queryString)
+
+  if (!response.ok) {
+    throw new Error(`Request failed: [${response.status}] ${response.body}`)
+  }
+
+  const orders: RawOrder[] = await response.json()
+
+  return orders
```
W3stside, line 223: dont want to just `return response.json()` ?

W3stside: uber nit that was hehe

W3stside (commented):
I am slightly concerned the non `try/catch` `async` methods could be a concern should they fail for reason(s) other than the `!response.ok`

src/api/operator/operatorApi.ts
alfetopito: Not sure, but yeah, good point. I thought I saw this somewhere before...
Here I'm building the URL query string (`?param=value&param=value`), and for sure there are things for building that already.
Will look it up

src/apps/explorer/components/OrderWidget/index.tsx
alfetopito: Yep yep, all of this will be rewritten

src/apps/explorer/components/OrderWidget/index.tsx
alfetopito: Yeah, it's definitely just for testing. Did not spend time on this component, right now it's just a place holder to display the details. Its time will come ;)

src/api/operator/operatorApi.ts
alfetopito: I definitely can, and I think I tried but typescript.
(Or did I? Don't know, will try again)

src/api/operator/index.ts
alfetopito: Because of Typescript.
Since not all implemented APIS are mocked and not all mocked APIs are implemented, there's a type mismatch.
I sold this by providing a default for either case, but TS doesn't like that.

src/api/operator/operatorApi.ts
alfetopito: try catching the hell out of them now

still throwing errors though, so it's up to the caller to decide what to do with them

I intend to wrap those calls into hooks that'll handle the errors

src/api/operator/operatorApi.ts
alfetopito: It worked \o/

src/api/operator/operatorApi.ts
```diff
@@ -163,15 +163,26 @@ export async function getOrder(params: GetOrderParams): Promise<RawOrder> {
...
+
+  try {
+    response = await _get(networkId, queryString)
+  } catch (e) {
+    const msg = `Failed to fetch ${queryString}`
+    console.error(msg, e.message)
+    throw new Error(msg)
+  }
 
   if (!response.ok) {
```
W3stside, line 176: I would move this to be under line 169 where you wait the response, and then throw if not ok

But fine either way

src/api/operator/operatorApi.ts
```diff
@@ -163,15 +163,26 @@ export async function getOrder(params: GetOrderParams): Promise<RawOrder> {
...
     throw new Error(`Request failed: [${response.status}] ${await response.text()}`)
```
W3stside, line 181: This await may also fail and will be uncaught but also probably fine

W3stside (approved):

W3stside:
Nice!!

src/api/operator/operatorApi.ts
alfetopito: Hmm, I see.
I wanted to keep it outside because they are 2 different failures:

The first is fail to execute the `fetch` network call, the second is a successful request with a non successful response.

And when I throw due to `!response.ok`, it would be captured by the `catch`

What makes me think whether we need error types to identify failure reasons (Error; Error400, Error500, etc)
But that might be overkill.

src/api/operator/operatorApi.ts
alfetopito: Yeah... that's true.
Will address it in another PR.

src/api/operator/operatorApi.ts
W3stside: what I would do, or i guess you could do is use inline `.catch` on `await` like:
```ts
// operatorApi.ts
...
try {
    response = await _get(networkId, queryString)
    if (!response.ok) {
      const responseText = await response.text().catch((e) => { throw 'Error getting response text' })
      throw `Request failed: [${response.status}] ${responseText}`)
    }
  } catch (e) {
    const msg = `Failed to fetch ${queryString}`
    console.error(msg, e.message)
    throw new Error(msg)
  }
...
```

src/api/operator/operatorApi.ts
W3stside: https://github.com/gnosis/gp-ui/pull/89#discussion_r566676968

src/apps/explorer/components/OrderWidget/index.tsx
anxolin: Yes, I think the design is not closed, so probably not too worth spending time on laying out. Although probably will be as david says...

src/api/operator/operatorApi.ts
alfetopito: Let's continue on https://github.com/gnosis/gp-ui/pull/93

PR #93, opened 2021-01-29
Addressing case where response.text() might throw

alfetopito:
From https://github.com/gnosis/gp-ui/pull/89#discussion_r566676968


src/api/operator/operatorApi.ts
```diff
@@ -178,13 +178,19 @@ export async function getOrder(params: GetOrderParams): Promise<RawOrder | null>
     if (response.status === 404) {
       return null
     }
-    throw new Error(`Request failed: [${response.status}] ${await response.text()}`)
+    // Just in case response.text() fails
+    const responseText = await response.text().catch((e) => {
```
W3stside, line 182: this will still be uncaught as it's not in the scope of a `try/catch` block, this is is why i recommended moving it up into the prvious `try/catch` block

```js
 let prom2 = await fakePromise(3000, true).catch(e => { throw new Error(e) })
// VM513:2 Uncaught Error: ERROR
//    at <anonymous>:1:63
//    at async <anonymous>:1
```

W3stside: just set the error message u want the `catch` block to use inside the inlined `throw new Error(<message>)` or just `throw "<message>"` and you can just have the `catch block` rethrow/log the error passed from the `try` throw

src/api/operator/operatorApi.ts
```diff
@@ -222,18 +228,24 @@ export async function getOrders(params: GetOrdersParams): Promise<RawOrder[]> {
...
     const msg = `Failed to fetch ${queryString}`
-    console.error(msg, e.message)
+    console.error(msg, e)
     throw new Error(msg)
   }
 
   if (!response.ok) {
-    throw new Error(`Request failed: [${response.status}] ${response.body}`)
+    // Just in case response.text() fails
+    const responseText = await response.text().catch((e) => {
```
W3stside, line 237: same as above, uncaught as it's not in a try/catch and the inlined `catch` is rethrowing

W3stside (changes requested):

src/api/operator/operatorApi.ts
anxolin: Well, I don't think is uncought. The function is asyn, so it just fails with the right message

anxolin (approved):
It feels a bit overcomplicated the error handling, but maybe is fine for now and we can just develop and improve as we touch it in the future.

Probably since we'll see some repetitive code in the error handling if we add more endpoints, and we'll likely want to move it to the `_get` function so `getOrder` just do `_get('/orders/${orderId}')`.

src/api/operator/operatorApi.ts
alfetopito: I want this to not be caught by the first try/catch. Besides, that statement also throws.
The only cases where it doesn't throw are when:
1. There is a valid response
2. The order doesn't exist (404) which then it returns `null`

Otherwise, it's a failure, and I want the caller to deal with it.

src/api/operator/operatorApi.ts
W3stside: It will read as uncaught in console as it's falling through the inline catch when it's rethrown in line 242. Solution is add it in try block or just leave as is. for me it's fine

W3stside (approved):

Issue #62, opened 2021-01-18
[view meta tx] Simple page and main layout and mocked data


anxolin:
# 🚨🚨🚨View order details

Requirements: 
- [ ] Allows to see the meta-transaction before is submitted in the settlement
  - [ ] Show all order fields
  - [x] Mock data on the page
- [ ] Search meta-transaction by ID
- [x] Open meta-transaction by URL
- [x] Status of the transaction: PENDING, EXECUTED, PARTIALLY_EXECUTED, EXPIRED, EXPIRED_PARTIALLY_EXECUTED?
- [ ] Link to settlement in the Explorer - if not PENDING nor EXPIRED
- [ ] Link to settlement transaction in Etherscan - if not PENDING nor EXPIRED

# Mockups 

https://www.figma.com/file/bEIQI9qESxqMVm5NkbtfZV/GP-Explorer-Wireframe-v1?node-id=1%3A6
(there's no detail view yet)

anxolin:
@alfetopito we can close and remove the undone things (move to other issues)

PR #97, opened 2021-01-29
62/order status

alfetopito:
# Summary

Part of #62 

Added logic for calculating order status.
Classifying it in one of:
1. Filled
2. Partially filled
3. Expired
4. Open

With unit tests.
Not yet added it to the UI. Will come in another PR


src/api/operator/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export type OrderStatus = 'open' | 'filled' | 'expired' | 'partially filled'
+
+// TODO: move to const or app settings?
+const FILLED_ORDER_EPSILON = '0.0001' // == 0.01%
+
+function isOrderFilled(order: RawOrder): boolean {
+  let amount, executedAmount
+
+  if (order.kind === 'buy') {
```
W3stside, line 17: do we have an `OrderType` enum somewhere we can use instead of string?

src/api/operator/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+    amount = new BigNumber(order.buyAmount)
```
W3stside, line 18: question: I wonder if `BigNumber.from(...)` is faster :thinking:

src/api/operator/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  return executedAmount.gte(minimumAmount)
+}
+
+function isOrderExpired(order: RawOrder): boolean {
+  return Math.floor(Date.now() / 1000) > order.validTo
+}
+
+function isOrderPartiallyFilled(order: RawOrder): boolean {
+  if (order.kind === 'buy') {
```
W3stside, line 35: i won't mention these again in this review but I think we should use an `OrderType` `enum`

src/api/operator/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+    return order.executedBuyAmount !== '0'
+  } else {
+    return order.executedSellAmount !== '0'
+  }
+}
+
+export function getOrderStatus(order: RawOrder): OrderStatus {
+  if (isOrderFilled(order)) {
+    return 'filled'
```
W3stside, line 44: also here would be nice for `OrderStatus` enum instead of the `type` above though I guess this one is fine due to autocomplete from function retrn type? wdyt @anxolin

src/api/operator/utils.ts
anxolin: I think we do. If it's not exported in the NPM package of the contracts, we should have the enum, or at least some constants

src/api/operator/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+import BigNumber from 'bignumber.js'
+
+import { ONE_BIG_NUMBER } from 'const'
+
+import { RawOrder } from './types'
+
+export type OrderStatus = 'open' | 'filled' | 'expired' | 'partially filled'
+
+// TODO: move to const or app settings?
+const FILLED_ORDER_EPSILON = '0.0001' // == 0.01%
```
anxolin, line 12: Yes, in the constants. I wouldn't put it in the settings. 

Actually, the settings is sth I'm not sure we have sorted out since this is multi-app thing. We don't necesarilly have the same setting file as Mesa. We can keep it simple for now

src/api/operator/utils.ts
anxolin: 💯

test/api/OperatorApi/orderStatus.test.ts
```diff
@@ -0,0 +1,208 @@
...
+    })
+    test('Filled, exact amount', () => {
+      const order: RawOrder = { ...BASE_ORDER, kind: 'buy', buyAmount: '100', executedBuyAmount: '100' }
+
+      expect(getOrderStatus(order)).toEqual('filled')
+    })
+    test('Filled, surplus', () => {
+      const order: RawOrder = { ...BASE_ORDER, kind: 'buy', buyAmount: '1000', executedBuyAmount: '1001' }
+
+      expect(getOrderStatus(order)).toEqual('filled')
```
anxolin, line 48: Can this case happen?

anxolin: Happy to be resilient with the case, but just curious

test/api/OperatorApi/orderStatus.test.ts
```diff
@@ -0,0 +1,208 @@
...
+    })
+  })
+})
+
+describe('Partially filled status', () => {
+  describe('Buy order', () => {
+    test('Not partially filled, fully filled', () => {
+      const order: RawOrder = { ...BASE_ORDER, kind: 'buy', buyAmount: '10000', executedBuyAmount: '10000' }
```
anxolin, line 78: This test is strange? right? u need to test what you expect, not what you "not-expect"
This is covered in the `Filled, exact amount` case

test/api/OperatorApi/orderStatus.test.ts
```diff
@@ -0,0 +1,208 @@
...
+
+      expect(getOrderStatus(order)).not.toEqual('partially filled')
+    })
+    test('Not partially filled, no fills', () => {
+      const order: RawOrder = { ...BASE_ORDER, kind: 'buy', buyAmount: '10000', executedBuyAmount: '0' }
+
+      expect(getOrderStatus(order)).not.toEqual('partially filled')
```
anxolin, line 85: same as above

test/api/OperatorApi/orderStatus.test.ts
```diff
@@ -0,0 +1,208 @@
...
+  beforeEach(mockTimes)
+
+  describe('Buy order', () => {
+    test('Not expired, open', () => {
+      const order: RawOrder = {
+        ...BASE_ORDER,
+        kind: 'buy',
+        buyAmount: '10000',
+        executedBuyAmount: '0',
+        validTo: Math.floor(Date.now() / 1000),
```
anxolin, line 122: maybe you can use a aux function `_getPastDate()` and `_getFutureDate()`? just to read the case easier?.

test/api/OperatorApi/orderStatus.test.ts
```diff
@@ -0,0 +1,208 @@
...
+    })
+    test('Filled, surplus', () => {
+      const order: RawOrder = { ...BASE_ORDER, kind: 'sell', sellAmount: '1000', executedSellAmount: '1001' }
+
+      expect(getOrderStatus(order)).toEqual('filled')
+    })
+  })
+})
+
+describe('Partially filled status', () => {
```
anxolin, line 75: I think you have missing the case where if the order is expired and partially filled, it's "Partially filled".

test/api/OperatorApi/orderStatus.test.ts
```diff
@@ -0,0 +1,208 @@
...
+  buyToken: '0xc7ad46e0b8a400bb3c915120d284aafba8fc4735',
+  validTo: 0,
+  appData: 0,
+  kind: 'sell',
+  partiallyFillable: false,
+  signature:
+    '0x04dca25f59e9ac744c4093530a38f1719c4e0b1ce8e4b68c8018b6b05fd4a6944e1dcf2a009df2d5932f7c034b4a24da0999f9309dd5108d51d54236b605ed991c',
+}
+
+describe('Filled status', () => {
```
anxolin, line 28: I think you have missing the case where if the order is expired and filled, it's "Filled".

test/api/OperatorApi/orderStatus.test.ts
```diff
@@ -0,0 +1,208 @@
...
+        ...BASE_ORDER,
+        kind: 'sell',
+        sellAmount: '10000',
+        executedSellAmount: '10',
+        validTo: Math.floor(Date.now() / 1000),
+      }
+      expect(getOrderStatus(order)).toEqual('open')
+    })
+  })
+})
```
anxolin, line 208: I think there's nice to add cases where you try that, "sellAmount" don't affect the state for a buy order (and the other way around for sell orders)

anxolin (commented):
Great PR! I really like your tests, they read super easy!

src/api/operator/utils.ts
alfetopito: Well, it's a type https://github.com/gnosis/gp-ui/blob/develop/src/api/operator/types.ts#L16

I feel like we had this discussion of types vs enums many months ago

src/api/operator/utils.ts
alfetopito: BigNumber does not have a static `.from()` method as far as I can tell https://mikemcl.github.io/bignumber.js/#bignumber

src/api/operator/utils.ts
alfetopito: Types do have autocomplete as well 
screenshot_2021-02-01_10-50-25

src/api/operator/utils.ts
anxolin: aha, okok, it would fail if it's anything different than buy, ok, then I'm fine with it

test/api/OperatorApi/orderStatus.test.ts
alfetopito: What I understood from backend guys is that this is possible.
In fact, good that you reminded me, I'll ask their input as well.

@bh2smith @nlordell @vkgnosis would you mind clarifying this question?
And while you are at it, also review the status logic on whether it matches how it works on the backend for both fill-or-kill and open orders?

src/api/operator/utils.ts
anxolin: If it's typed, It's fine for me as it's done here

test/api/OperatorApi/orderStatus.test.ts
anxolin: I just think is weird that someone would receive more tokens than he ask in a buy order. 
If I get a better price, then I would like to spend less "sell token"

test/api/OperatorApi/orderStatus.test.ts
alfetopito: Yes I agree it looks weird.
I added this kind of tests as a way of testing each individual status checking function only through the general one.

Meaning, since I do not export `isFilledOrder` and so on, I still split the tests per status, and test all possible cases for each, even though it might have been covered in other step.

alfetopito: Refactored tests to have only expected on each section

test/api/OperatorApi/orderStatus.test.ts
alfetopito: Actually, `partially filled` status is only set when the order is expired, otherwise it's `open`.
So in fact all `partially filled` orders are expired.

test/api/OperatorApi/orderStatus.test.ts
alfetopito: All orders are expired in this unittest, so I was missing the other way around (filled and NOT expired).
Added.

test/api/OperatorApi/orderStatus.test.ts
alfetopito: Done

src/api/operator/utils.ts
alfetopito: Like the other function, this also has auto-complete and TS knows it refers to a type, not a string:
screenshot_2021-02-01_13-56-53

test/api/OperatorApi/orderStatus.test.ts
nlordell: This cannot happen. The total executed buy amount can not exceed the order's buy amount. A surplus in a buy order gets translated to `totalExecutedSellAmount < sellAmount`.

One way to think of this, is that for a buy order you always buy exactly what you asked for, you may just pay less that what you said your maximum was.

src/api/operator/utils.ts
```diff
@@ -0,0 +1,51 @@
...
+
+function isOrderFilled(order: RawOrder): boolean {
+  let amount, executedAmount
+
+  if (order.kind === 'buy') {
+    amount = new BigNumber(order.buyAmount)
+    executedAmount = new BigNumber(order.executedBuyAmount)
+  } else {
+    amount = new BigNumber(order.sellAmount)
+    executedAmount = new BigNumber(order.executedSellAmount)
```
nlordell, line 19: For sell orders, this should be:
```suggestion
    executedAmount = new BigNumber(order.executedSellAmount).minus(order.executedFeeAmount)
```

This is because in the SC we wanted the trade event to emit the exact amount that was being transferred from the user, and I suspect the back end just returns this value without modification. @vkgnosis brought this up in an issue to change the event (which would also change the value returned by the back end) to have this changed to be the filled sell amount: https://github.com/gnosis/gp-v2-contracts/issues/392.

test/api/OperatorApi/orderStatus.test.ts
```diff
@@ -0,0 +1,248 @@
...
+
+      expect(getOrderStatus(order)).toEqual('filled')
+    })
+    test('Filled, exact amount', () => {
+      const order: RawOrder = { ...ORDER, kind: 'sell', sellAmount: '100', executedSellAmount: '100' }
+
+      expect(getOrderStatus(order)).toEqual('filled')
+    })
+    test('Filled, surplus', () => {
+      const order: RawOrder = { ...ORDER, kind: 'sell', sellAmount: '1000', executedSellAmount: '1001' }
```
nlordell, line 71: This is only possible if `feeAmount >= 1`

test/api/OperatorApi/orderStatus.test.ts
```diff
@@ -0,0 +1,248 @@
...
+      expect(getOrderStatus(order)).toEqual('filled')
+    })
+    test('Filled, sell amount does not affect output', () => {
+      const order: RawOrder = {
+        ...ORDER,
+        kind: 'buy',
+        buyAmount: '100',
+        executedBuyAmount: '100',
+        sellAmount: '100',
+        executedSellAmount: '1100',
```
nlordell, line 53: This is only possible if `feeAmount >= 100`

src/api/operator/utils.ts
anxolin: I wrote in the issue, but in short I think the current approach is ok, although I have no super strong opinion about that. Is just good to know and be explicit about it. Thanks for pointing this out

test/api/OperatorApi/orderStatus.test.ts
anxolin: Agree, although I would leave the test, but maybe rename it to `Filled, surplus (not really possible in contracts)`. I mean, the alternative would be to be to throw in these cases, and it feel unnecessarily strict. It's nice to say in these case the order is filled.

anxolin: If we insist, I wouldn't be opposed to throw, but I think is just too much for the states to add these validations. I think is enough to be resilient to these.

anxolin (approved):
Only blocker of this PR in my eyes is Nicks comment on the sellAmount. Other than that, good to go.

test/api/OperatorApi/orderStatus.test.ts
nlordell: To be clear, I wasn't suggesting that the code was handling this incorrectly, it was more about the fact that the test isn't really testing a "filled with surplus" state, what would be something like:
```ts
{
  ...ORDER,
  kind: 'sell',
  sellAmount: '1000',
  executedSellAmount: '1000',
  buyAmount: '1000',
  executedBuyAmount: '1337',
}
```

So maybe the test should be named something along the lines of `accepts invalid over-executed trades`?

test/api/OperatorApi/orderStatus.test.ts
nlordell: Same as :point_down:, no issues against the test besides that I think its incorrectly named. A completely filled buy order with a surplus would look like:
```ts
{
  ...ORDER,
  kind: 'sell',
  buyAmount: '1000',
  executedBuyAmount: '1000',
  sellAmount: '1000',
  executedSellAmount: '777',
  feeAmount: '0',
}
```

test/api/OperatorApi/orderStatus.test.ts
alfetopito: Since the surplus does not affect at all the order status, I'm in favor or removing surplus tests (besides the fact that they represent an invalid state on the backed)

src/api/operator/utils.ts
W3stside: samesies

src/api/operator/utils.ts
alfetopito: Updated and added unit tests accounting for the fee

alfetopito:
O shit, mergify nooooo.

Oh well, will send another PR

alfetopito:
Follow up https://github.com/gnosis/gp-ui/pull/103

anxolin:
Ahahah, Mergify. I think it's fine! then new PR with the changes is smaller and easier to review :P

PR #102, opened 2021-02-02
Table cleanup + styles

biocom:
- Simplified CSS in child component, by moving to simpleTable parent component.
- Integrated styles according to latest design.
- Responsive table.

Screen Shot 2021-02-02 at 18 25 26
Screen Shot 2021-02-02 at 18 25 08


src/components/common/SimpleTable/index.tsx
```diff
@@ -85,11 +101,6 @@ const Wrapper = styled.table<{ $numColumns?: number }>`
       display: flex;
       align-items: center;
       justify-content: flex-end;
-
-      &:first-of-type {
-        padding-left: 1rem;
-        justify-content: flex-start;
-      }
```
alfetopito, line 92: First column no longer aligns left then:
screenshot_2021-02-02_13-05-22
Storybook https://pr102--gpui.review.gnosisdev.com/storybook/?path=/docs/common-simpletable--custom-table-style
Trade app https://pr102--gpui.review.gnosisdev.com/trade.html

Is that ok?

alfetopito (approved):
Only one comment regarding the first column justification, otherwise looks great :+1:

src/components/common/SimpleTable/index.tsx
biocom: Good call, forgot that simpleTable is used by this component in the classic interface. Will fix.

src/apps/explorer/components/OrderWidget/OrderDetails.tsx
```diff
@@ -8,36 +8,25 @@ import { RawOrder } from 'api/operator'
 import { SimpleTable } from 'components/common/SimpleTable'
 
 // TODO: move to theme AAAAND pick color for white variant
-const COLUMN_COLOR = '#252535'
+const FONT_COLOR_DESCRIPTION = 'rgba(255, 255, 255, 0.9);';
+const TABLE_BORDER_COLOR = 'rgba(151, 151, 184, 0.3)';
```
W3stside, line 12: we should avoid mixing too many floating constants and theme... try to use the theme whereever possible please

src/apps/explorer/components/OrderWidget/OrderDetails.tsx
```diff
@@ -46,10 +35,8 @@ const Table = styled(SimpleTable)`
         }
       }
 
-      height: 5rem;
-
-      :not(:first-of-type) {
-        padding-left: 2.5rem;
+      &:last-of-type {
+        color: ${FONT_COLOR_DESCRIPTION};
```
W3stside, line 39: same here, theme

src/components/common/SimpleTable/index.tsx
```diff
@@ -1,14 +1,19 @@
 import React from 'react'
 import styled from 'styled-components'
+import { MEDIA } from 'const'
+
+// TODO: move to theme AAAAND pick color for white variant
+const FONT_COLOR = '#FFFFFF';
```
W3stside, line 6: unless we expect the Table to always have these styles, we need to use theme in these places. the earlier the better else in the future we'll need a larger rework. to make it simple just add this colour scheme to all themes or sth

src/components/common/SimpleTable/index.tsx
```diff
@@ -50,19 +56,30 @@ const Wrapper = styled.table<{ $numColumns?: number }>`
...
+      transition: background-color 0.1s ease-in-out;
+      height: ${CELL_HEIGHT};
+      padding: 1rem;
+      box-sizing: border-box;
+
+      &:not(:last-of-type) {
+        border-bottom: 0.1rem solid ${TR_BORDER_COLOR};
+      }
+
+      @media ${MEDIA.mobile} {
```
W3stside, line 68: do we not have the `theme.upToSmall` for this?

W3stside (changes requested):

anxolin (commented):
I'm good with the changes, only minor concerns what was already mentioned by david and leandro. 

Also, although for the  `OrderWidget` shows alright, for the Table example, in responsive doesn't look so good. 

What should be the default for the table?

image

src/components/common/SimpleTable/index.tsx
biocom: @W3stside Updated to use that, after help from @alfetopito . I do have some thoughts on that approach. The new approach:

```
      ${applyMediaStyles('upToSmall')`
        display: flex;
        flex-flow: column wrap;
        height: auto;
        align-items: flex-start;

        &:hover {
          background: var(--color-text-hover);
          > td {
            color: var(--color-text-primary);
          }
        }
      `}
```

Whereas before we had:
```
@media ${MEDIA.mobile} {
    ..CSS HERE..
}
```

In my view, the previous method was more elegant:
- Respect native (S)CSS. 
- Allows for normal syntax highlighting (because of native (S)CSS).
- Simple use of @media + a variable (could still come from our theme as it did).

The new approach is so much more verbose and harder to read/debug IMO, because of using a string literal for the code:
Screen Shot 2021-02-09 at 17 12 11

Your call otherwise.

src/components/common/SimpleTable/index.tsx
biocom: Should be fixed. As for fine tuning the classic UI table, I'd suggest we address it in a separate PR.

src/components/common/SimpleTable/index.tsx
biocom: Fixed.

src/apps/explorer/components/OrderWidget/OrderDetails.tsx
biocom: Fixed.

src/apps/explorer/components/OrderWidget/OrderDetails.tsx
biocom: Moved to using theme vars.

biocom:
> Also, although for the `OrderWidget` shows alright, for the Table example, in responsive doesn't look so good.
> 
> What should be the default for the table?

Couldn't fully test because of issues in the Storybook. Please let's address in separate PR.

src/apps/explorer/components/OrderWidget/OrderDetails.tsx
anxolin: @biocom I marked some as fixed. We usually mark as resolved ourselves if we clearly address the issue

anxolin (approved):

alfetopito:
Throwing the towel for now.
Spent almost the whole day trying to get the storybook working again, and failed.

In summary, the global state is not being passed down properly to the storybook.

Did not get to check whether the tests are failing because of that or not. Haven't even noticed until now to be honest.
Hmm, there might be another reason why this is failing then?

Anyway, I bisected the PR and it broke after e0c4fd3d27aabd1fb7113adfe219a01ae1b38c30
Even though I'm almost sure I tested when David made that change and it was working fine. Actually, the point of that change was to add styles to storybook :shrug:

alfetopito:
I checked the build and it was a 500 from github. Restarted and it's all green now.
Will merge and leave the storybook fix for another time

PR #105, opened 2021-02-03
62/order util fns

alfetopito:
# Summary

Part of #62 

Small util functions to be used when dealing with orders
- `getOrderFilledAmount` 
  Returns filed amount per order both as raw amount and as a percentage. Will be useful for the percentage and filled part of:
screenshot_2021-02-03_13-44-08
  Has unit tests

- `getOrderExecutedAmounts`
  Returns both buy/sell executed amounts as BigNumber. Will be useful anywhere we need to use executed amounts (the right most value in the picture above)
  Syntactic sugar because `executedSellAmount` is derived
  No unit tests (too dumb to deserve one)

- `getOrderLimitPrice`
  Returns order limit price.
  Optionally returns the price inverted:
screenshot_2021-02-03_14-55-08
  Has unit tests

- `getOrderExecutedPrice`
  Same as the previous, except that it returns the price for executed amounts. 
  Not in the UI AFAICT, but I'm sure we'll need it.
  Has unit tests

### DIFF
--- a/src/api/operator/utils.ts
+++ b/src/api/operator/utils.ts
@@ -3,7 +3,7 @@ import BigNumber from 'bignumber.js'
 
 import { calculatePrice, invertPrice } from '@gnosis.pm/dex-js'
 
-import { FILLED_ORDER_EPSILON, ONE_BIG_NUMBER } from 'const'
+import { FILLED_ORDER_EPSILON, ONE_BIG_NUMBER, ZERO_BIG_NUMBER } from 'const'
 
 import { OrderStatus, RawOrder } from './types'
 
@@ -131,6 +131,12 @@ export function getOrderExecutedPrice({
   inverted,
 }: GetOrderPriceParams): BigNumber {
   const { executedBuyAmount, executedSellAmount } = getOrderExecutedAmounts(order)
+
+  // Only calculate the price when both values are set
+  // Having only one value > 0 is anyway an invalid state
+  if (executedBuyAmount.isZero() || executedSellAmount.isZero()) {
+    return ZERO_BIG_NUMBER
+  }
 
   const price = calculatePrice({
     numerator: { amount: executedBuyAmount, decimals: buyTokenDecimals },
--- a/src/api/operator/utils.ts
+++ b/src/api/operator/utils.ts
@@ -65,9 +65,7 @@ export function getOrderFilledAmount(order: RawOrder): { amount: BigNumber; perc
     totalAmount = new BigNumber(order.sellAmount)
   }
 
-  const percentage = executedAmount.isZero() ? executedAmount : executedAmount.div(totalAmount)
-
-  return { amount: executedAmount, percentage }
+  return { amount: executedAmount, percentage: executedAmount.div(totalAmount) }
 }
 
 /**
--- a/src/api/operator/utils.ts
+++ b/src/api/operator/utils.ts
@@ -1,10 +1,10 @@
 // Util functions that only pertain to/deal with operator API related stuff
-
 import BigNumber from 'bignumber.js'
 
-import { FILLED_ORDER_EPSILON, ONE_BIG_NUMBER } from 'const'
+import { calculatePrice, invertPrice } from '@gnosis.pm/dex-js'
 
 import { RawOrder } from './types'
+import { FILLED_ORDER_EPSILON, ONE_BIG_NUMBER } from 'const'
 
 export type OrderStatus = 'open' | 'filled' | 'expired' | 'partially filled'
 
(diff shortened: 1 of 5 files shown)


src/api/operator/utils.ts
```diff
@@ -49,3 +48,100 @@ export function getOrderStatus(order: RawOrder): OrderStatus {
...
+
+/**
+ * Get order filled amount, both as raw amount (in atoms) and as percentage (from 0 to 1)
+ *
+ * @param order The order
+ */
+export function getOrderFilledAmount(order: RawOrder): { amount: BigNumber; percentage: BigNumber } {
+  let executedAmount, totalAmount
+
+  if (order.kind === 'buy') {
```
W3stside, line 60: dont' we need to add the fee to the selling amount when it's a buy?

src/api/operator/utils.ts
```diff
@@ -49,3 +48,100 @@ export function getOrderStatus(order: RawOrder): OrderStatus {
...
+    executedAmount = new BigNumber(order.executedBuyAmount)
+    totalAmount = new BigNumber(order.buyAmount)
+  } else {
+    executedAmount = new BigNumber(order.executedSellAmount).minus(order.executedFeeAmount)
```
W3stside, line 64: since you're subtracting it here it only seems logical but maybe I'm missing sth

src/api/operator/utils.ts
```diff
@@ -49,3 +48,100 @@ export function getOrderStatus(order: RawOrder): OrderStatus {
...
+    totalAmount = new BigNumber(order.sellAmount)
+  }
+
+  const percentage = executedAmount.isZero() ? executedAmount : executedAmount.div(totalAmount)
```
W3stside, line 68: is it safe to divide from zero in `BigNumber`? then this could just be 
```ts
const percentage = executedAmount.div(totalAmount)
```
as it will still be zero

W3stside (approved):
super nice, clean, easy to read. :rocket:

src/api/operator/utils.ts
alfetopito: It's safe, but this is to avoid returning NaN or Infinity
When we know it's 0, just return 0 and avoid issues later on when displaying it

src/api/operator/utils.ts
alfetopito: I don't get the question :confused:

src/api/operator/utils.ts
alfetopito: On both order types (sell & buy) the fee must be taken into account, but only for executed sell amounts.
That's because we only take the fee in the sell token.
So, executed buy amounts are indifferent to the fee

src/api/operator/utils.ts
alfetopito: Well actually, I ran the tests with your suggestion and it still works.
Either I don't test the case I had in mind or it works as you describe :)

Will update

src/api/operator/utils.ts
```diff
@@ -49,3 +48,98 @@ export function getOrderStatus(order: RawOrder): OrderStatus {
     return 'open'
   }
 }
+
+/**
+ * Get order filled amount, both as raw amount (in atoms) and as percentage (from 0 to 1)
+ *
+ * @param order The order
+ */
+export function getOrderFilledAmount(order: RawOrder): { amount: BigNumber; percentage: BigNumber } {
```
anxolin, line 57: why is not asking you to add the return type? did we turn off the super strict mode?

anxolin (approved):
Awesome. Not need to do for now, but probably this utils don't belong in the `api/operator`, they look to me pretty general. I see some of them use the "RawData" in the input, so I think it's fine for now. Just, if we see it using it a lot in other places we should promote it to `utils/` 

anyways, awesome tests!

src/api/operator/utils.ts
alfetopito: What do you mean? The return type is there :thinking:

alfetopito:
> Awesome. Not need to do for now, but probably this utils don't belong in the `api/operator`, they look to me pretty general. I see some of them use the "RawData" in the input, so I think it's fine for now. Just, if we see it using it a lot in other places we should promote it to `utils/`

My idea to keep these utils within `/operatorApi` is because they are the functional equivalent to class methods.
These functions operate on the return of the API, and do not stand on their own.

Would potentially be part of the API itself, but since not directly related to API calls, they're on a separated file.

In summary: these fns are part of the API, but not pure API fns.

Maybe the file name (`utils`) is misleading?

> anyways, awesome tests!

:partying_face:

anxolin:
sounds good!

PR #115, opened 2021-02-04
66/split global state

### LINKED ISSUE #66: [view meta tx] Global State
- One state per app
- Stores are the reusable part
- Use them store

alfetopito:
# Summary

Closes #66 

Split global state per app

Main changes:
- `app/state.ts` contains all state definitions/reducers/etc. Same that was before defined on `reducers-actions`
- Some duplication might be required on each app's `state.ts` definition
- `useGlobalState` now requires the state type to tell TS what type is your global state holding. Example:
screenshot_2021-02-04_13-33-50
- For now only `theme` state is part of the global state

Let me know what you think about this approach.

### DIFF
--- a/src/hooks/useGlobalState.tsx
+++ b/src/hooks/useGlobalState.tsx
@@ -1,13 +1,12 @@
 import React, { useContext, useReducer } from 'react'
-import { GlobalState } from 'reducers-actions'
 import { AnyAction } from 'combine-reducers'
 
 const GlobalStateContext = React.createContext({})
 
-export function withGlobalContext<P>(
+export function withGlobalContext<P, State>(
   WrappedComponent: React.FC<P>,
-  initialStateFunc: () => GlobalState,
-  reducer: React.Reducer<GlobalState, AnyAction>,
+  initialStateFunc: () => State,
+  reducer: React.Reducer<State, AnyAction>,
 ): (props: P) => JSX.Element {
   return function WrappedComponentWithGlobalState(props: P): JSX.Element {
     const [state, dispatch] = useReducer(reducer, initialStateFunc())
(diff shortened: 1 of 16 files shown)


anxolin (commented):
I very much like the changes in the PR. 

This is an approved, for this PR, but I'd like you to consider the comments I mentioned, and also for other PRs, the state I think is a bit messy. 

I suggest following the structure like Uniswap, that's easy to understand and define slices of state we can reuse in apps. 

image

This would imply to rename `reducers-actions` to `state`, and break the file into `actions` and `reducer`.  I think we don't have updater, for now, and for v1 we wouldn't need to adapt it into that for now. But nice to have the concept in mind in case we want to do the same thing. 

Regarding `addSideEffect`, I'm not sure why we went for that approach, I find it a bit of boilerplate and engineered.   I believe `redux-thunk`,  `redux-promise` are more standard ways of dealing with a similar problem. Although I feel that probably having hooks, we just need to move side effects to the hooks and make reducers more pure and synchronous.

Maybe @W3stside has an opinion in these side effects. Mi feeling right now, is let's create hooks near the state in a similar way to uni.

anxolin (approved):

src/hooks/useGlobalState.tsx
```diff
@@ -1,13 +1,12 @@
 import React, { useContext, useReducer } from 'react'
-import { GlobalState } from 'reducers-actions'
 import { AnyAction } from 'combine-reducers'
 
 const GlobalStateContext = React.createContext({})
 
-export function withGlobalContext<P>(
+export function withGlobalContext<P, State>(
```
W3stside, line 6: lol nice, fixing my shitty types

W3stside (approved):

src/hooks/useGlobalState.tsx
alfetopito: Nothing really wrong with your type, thinking that there shouldn't be more than 1 global state.

PR #104, opened 2021-02-02
62/order status component

alfetopito:
# Summary

Part of #62

Implemented order status component as per [design](https://www.figma.com/file/zcWhLymqovrJ4Lj1v9FoSI/GP-Explorer?node-id=0%3A1):
screenshot_2021-02-02_15-09-50

# Testing
Check orderbook https://pr104--gpui.review.gnosisdev.com/storybook/?path=/docs/order-statuslabel--expired

PR #116, opened 2021-02-04
Refactoring CSS logic to external function

alfetopito:
Part of #62 

Follow up to comment on #104: https://github.com/gnosis/gp-ui/pull/104#discussion_r570513618

2 equivalent approaches for the refactoring, shown together:
screenshot_2021-02-04_14-03-56

This change for now (commit 8de77dc) uses the first, returning CSS from the helper function

### DIFF
--- a/src/components/order/StatusLabel/index.tsx
+++ b/src/components/order/StatusLabel/index.tsx
@@ -42,7 +42,7 @@ const Wrapper = styled.div<Props>`
   padding: 0.75em;
   display: inline-block;
 
-  ${(...params): string => setStatusColors(...params)}
+  ${({ theme, status }): string => setStatusColors({ theme, status })}
 `
 
 const StyledFAIcon = styled(FontAwesomeIcon)`
--- a/src/components/order/StatusLabel/index.tsx
+++ b/src/components/order/StatusLabel/index.tsx
@@ -1,11 +1,36 @@
 import React from 'react'
-import styled from 'styled-components'
+import styled, { DefaultTheme } from 'styled-components'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { faCheckCircle, faClock, faDotCircle, IconDefinition } from '@fortawesome/free-solid-svg-icons'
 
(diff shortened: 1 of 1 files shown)


anxolin (approved):
Great Thankssss! I feel this is much better. I gave in with writing CSS in JS, but at least for this big functions we could try not to mix it

src/components/order/StatusLabel/index.tsx
```diff
@@ -1,12 +1,37 @@
...
 import { OrderStatus } from 'api/operator'
 
 export type Props = { status: OrderStatus }
 
+function setStatusColors({ theme, status }: { theme: DefaultTheme; status: OrderStatus }): string {
+  let background, text
+
+  switch (status) {
+    case 'expired':
+      text = theme.labelTextExpired
```
W3stside, line 15: why not `color`? haha super nit time

src/components/order/StatusLabel/index.tsx
```diff
@@ -17,29 +42,7 @@ const Wrapper = styled.div<Props>`
...
-        color = theme.labelTextOpen
-        background = theme.labelBgOpen
-        break
-    }
-    return `
-      background: ${background};
-      color: ${color};
-    `
-  }}
+  ${(...params): string => setStatusColors(...params)}
```
W3stside, line 45: hmm i think you\re unnecessarily passing too many arguments here, probably better to destructure as normal?

W3stside (commented):
minor comments then can approve

src/components/order/StatusLabel/index.tsx
alfetopito: Fun story, that's what I had.
But both are colors (background and text) to I would either prefix both or none. I chose to keep only the name of the css property.

src/components/order/StatusLabel/index.tsx
alfetopito: Just wanted to reduce the logic in the CSS as Anxo suggested.
But, no problem, will update.

PR #117, opened 2021-02-05
62/erc20 global state

alfetopito:
# Summary

Part of #62 

Implementing `Erc20` state manager using reducer style.
Added hook `useErc20` to get erc20 from global context or fetch from the network.

Right now the hook is plugged into OrderWidget and logging the sell/buy token on the console, but that's not where it's intended to be used. Will be addressed in a future PR.

### DIFF
--- a/src/apps/explorer/api.ts
+++ b/src/apps/explorer/api.ts
@@ -1,1 +1,9 @@
-export { web3 } from 'api'
+import { createErc20Api } from 'api'
+
+import { web3 } from 'api'
+
+const injectedDependencies = { web3 }
+
+const erc20Api = createErc20Api(injectedDependencies)
+
+export { web3, erc20Api }
--- a/src/apps/explorer/components/OrderWidget/index.tsx
+++ b/src/apps/explorer/components/OrderWidget/index.tsx
@@ -4,6 +4,7 @@ import { useParams } from 'react-router'
 import { getOrder, RawOrder } from 'api/operator'
 
 import { OrderWidgetView } from './view'
+import { useErc20 } from 'hooks/useErc20'
 
 export const OrderWidget: React.FC = () => {
   // TODO: move order loading to a hook
@@ -26,6 +27,13 @@ export const OrderWidget: React.FC = () => {
       .catch((e) => setError(e.message))
       .finally(() => setIsLoading(false))
   }, [orderId])
+
+  // TODO: this is just for testing. The hooks will not be here
+  const { value: buyToken } = useErc20(order?.buyToken, 4)
+  const { value: sellToken } = useErc20(order?.sellToken, 4)
+
+  console.log(`buy token`, buyToken)
+  console.log(`sell token`, sellToken)
 
   return <OrderWidgetView order={order} isLoading={isLoading} error={error} />
 }
--- a/src/apps/explorer/components/OrderWidget/index.tsx
+++ b/src/apps/explorer/components/OrderWidget/index.tsx
@@ -29,8 +29,8 @@ export const OrderWidget: React.FC = () => {
   }, [orderId])
 
   // TODO: this is just for testing. The hooks will not be here
-  const { value: buyToken } = useErc20(order?.buyToken, 4)
-  const { value: sellToken } = useErc20(order?.sellToken, 4)
+  const { value: buyToken } = useErc20({ address: order?.buyToken, networkId: 4 })
+  const { value: sellToken } = useErc20({ address: order?.sellToken, networkId: 4 })
 
   console.log(`buy token`, buyToken)
   console.log(`sell token`, sellToken)
--- a/src/hooks/useErc20.ts
+++ b/src/hooks/useErc20.ts
@@ -14,10 +14,16 @@ import { web3, erc20Api } from 'apps/explorer/api'
 
 import { ExplorerAppState } from 'apps/explorer/state'
 
-export function useErc20(
-  address?: string,
-  networkId?: Network,
-): { isLoading: boolean; error?: string; value?: TokenErc20 } {
+type Params = {
+  address?: string
+  networkId?: Network
+}
+
+type Return = { isLoading: boolean; error?: string; value?: TokenErc20 }
+
+export function useErc20(params: Params): Return {
+  const { address, networkId } = params
+
   const [isLoading, setIsLoading] = useState(false)
   const [error, setError] = useState('')
 
--- a/src/hooks/useErc20.ts
+++ b/src/hooks/useErc20.ts
@@ -1,0 +1,59 @@
+import { useEffect, useState } from 'react'
+
+import { TokenErc20 } from '@gnosis.pm/dex-js'
+
+import { Network } from 'types'
+
+import { buildErc20Key, saveSingleErc20 } from 'reducers-actions/erc20'
+
+import useGlobalState from 'hooks/useGlobalState'
+
+import { getErc20Info } from 'services/helpers'
+
+import { web3, erc20Api } from 'apps/explorer/api'
+
+import { ExplorerAppState } from 'apps/explorer/state'
+
+export function useErc20(
+  address?: string,
+  networkId?: Network,
+): { isLoading: boolean; error?: string; value?: TokenErc20 } {
+  const [isLoading, setIsLoading] = useState(false)
+  const [error, setError] = useState('')
+
+  const [{ erc20s }, dispatch] = useGlobalState<ExplorerAppState>()
+
+  // Initial state from global state
+  const [erc20, setErc20] = useState<TokenErc20 | undefined>(
+    address && networkId ? erc20s.get(buildErc20Key(networkId, address)) : undefined,
+  )
+
+  useEffect(() => {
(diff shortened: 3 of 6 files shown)


src/apps/explorer/components/OrderWidget/index.tsx
```diff
@@ -27,5 +28,12 @@ export const OrderWidget: React.FC = () => {
       .finally(() => setIsLoading(false))
   }, [orderId])
 
+  // TODO: this is just for testing. The hooks will not be here
+  const { value: buyToken } = useErc20(order?.buyToken, 4)
+  const { value: sellToken } = useErc20(order?.sellToken, 4)
```
W3stside, line 33: can this take object params? pls

src/hooks/useErc20.ts
```diff
@@ -0,0 +1,59 @@
...
+import { getErc20Info } from 'services/helpers'
+
+import { web3, erc20Api } from 'apps/explorer/api'
+
+import { ExplorerAppState } from 'apps/explorer/state'
+
+export function useErc20(
+  address?: string,
+  networkId?: Network,
+): { isLoading: boolean; error?: string; value?: TokenErc20 } {
```
W3stside, line 20: nit: make into an interface

src/hooks/useErc20.ts
```diff
@@ -0,0 +1,59 @@
...
+  const [isLoading, setIsLoading] = useState(false)
+  const [error, setError] = useState('')
+
+  const [{ erc20s }, dispatch] = useGlobalState<ExplorerAppState>()
+
+  // Initial state from global state
+  const [erc20, setErc20] = useState<TokenErc20 | undefined>(
```
W3stside, line 27: does `undefined` need to be explicit here?

src/hooks/useErc20.ts
```diff
@@ -0,0 +1,59 @@
...
+    address && networkId ? erc20s.get(buildErc20Key(networkId, address)) : undefined,
```
W3stside, line 28: oh i see

src/reducers-actions/erc20.ts
```diff
@@ -0,0 +1,48 @@
...
+export function reducer(state: Erc20State, action: ReducerActionType): Erc20State {
+  switch (action.type) {
+    case 'SAVE_MULTIPLE_ERC20': {
+      // Clone current state
+      const map = new Map(state)
+
+      const { erc20s, networkId } = action.payload
+
+      erc20s.forEach((erc20) => {
+        // add/overwrite details for given key
```
W3stside, line 39: do we want override and not caching?

W3stside (approved):
approving with minor comments. nice work!

src/apps/explorer/components/OrderWidget/index.tsx
alfetopito: Sure, will do

src/hooks/useErc20.ts
alfetopito: Good call

src/hooks/useErc20.ts
alfetopito: Yes, otherwise TS complains

src/reducers-actions/erc20.ts
alfetopito: Well, in this case (erc20 token details, with 4 fields) I don't really care what's there, because there shouldn't be anything there to begin with.
If all goes according to plan (:crossed_fingers:), the action to persist a token will only be called when it's not found in the global state in the first place.

Anyway, I will change to merge existing properties, just didn't consider this is something that should happen.

alfetopito:
@anxolin , will merge this now to avoid painful conflicts with https://github.com/gnosis/gp-ui/pull/129.
Will address any comments you might have as new PRs

src/hooks/useErc20.ts
```diff
@@ -0,0 +1,65 @@
+import { useEffect, useState } from 'react'
```
anxolin, line 1: I like more the structure of Uniswap, where for this kind of hooks they have them inside of state.

In this case would be `state/erc20/hooks.ts`

src/hooks/useErc20.ts
```diff
@@ -0,0 +1,65 @@
...
+
+type Return = { isLoading: boolean; error?: string; value?: TokenErc20 }
+
+export function useErc20(params: Params): Return {
+  const { address, networkId } = params
+
+  const [isLoading, setIsLoading] = useState(false)
+  const [error, setError] = useState('')
+
+  const [{ erc20s }, dispatch] = useGlobalState<ExplorerAppState>()
```
anxolin, line 30: I know we did it like this in the past, but why not using `useSelector` https://react-redux.js.org/api/hooks

src/hooks/useErc20.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  // Initial state from global state
+  const [erc20, setErc20] = useState<TokenErc20 | undefined>(
+    address && networkId ? erc20s.get(buildErc20Key(networkId, address)) : undefined,
+  )
+
+  useEffect(() => {
+    // Only try to fetch it if not on global state
+    if (!erc20 && address && networkId) {
```
anxolin, line 39: I would try to reduce the complexity of this functions

Can we just, for example, instead of wrapping all in a if, do a `if(the opposite) return undefined`
Then u reduce all the cyclomatic complexity by 1

Not sure if it's worth it, but we can use async/await syntax. For example:

```ts
useEffect(() => {
    async function getErc20AndPersist(): Promise<void> {
      // Only try to fetch it if not on global state
      if (erc20 || !address || !networkId) {
        return undefined
      }

      try {
        const fetchedErc20 = await getErc20Info({ tokenAddress: address, networkId, web3, erc20Api })

        // When not found, it returns null
        if (fetchedErc20) {
          dispatch(saveSingleErc20(fetchedErc20, networkId))
          setErc20(fetchedErc20)
        }
      } catch (error) {
        // Set error only when the call fails
        // Not finding the token is not an error
        const msg = `Failed to fetch erc20 details for ${address} on network ${networkId}`
        console.error(msg, error)
        setError(msg)
      } finally {
        setIsLoading(false)
      }
    }

    getErc20AndPersist()
  }, [address, dispatch, erc20, networkId])
```

src/reducers-actions/erc20.ts
```diff
@@ -0,0 +1,50 @@
+import { TokenErc20 } from '@gnosis.pm/dex-js'
```
anxolin, line 1: I'd love to follow the structure of uni, I like how they separate have all slices of the state in folders, and they split actions, hooks, redicers ...

anxolin (commented):
APPROVED, I just added some comments

anxolin:
Good call merging it already! I can always review after. If I take to much time, u can always merge or tell me (or both)

src/apps/explorer/api.ts
```diff
@@ -1 +1,9 @@
-export { web3 } from 'api'
+import { createErc20Api } from 'api'
```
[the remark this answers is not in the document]
alfetopito: That makes sense, each app has it's own API index. Will create a new PR for that.

src/hooks/useErc20.ts
```diff
@@ -0,0 +1,65 @@
...
+import { Network } from 'types'
+
+import { buildErc20Key, saveSingleErc20 } from 'reducers-actions/erc20'
+
+import useGlobalState from 'hooks/useGlobalState'
+
+import { getErc20Info } from 'services/helpers'
+
+import { web3, erc20Api } from 'apps/explorer/api'
+
```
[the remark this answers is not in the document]
alfetopito: Well, I like to split the imports by groups. This time there are a LOT of imports, but couldn't group them together.

src/hooks/useErc20.ts
alfetopito: Check https://github.com/gnosis/gp-ui/pull/130

If I get your point, in that PR all erc20 related hooks should be on /erc20/hooks then?
I'll wait for your input there.

src/reducers-actions/erc20.ts
alfetopito: That's on this PR :) https://github.com/gnosis/gp-ui/pull/129

src/apps/explorer/api.ts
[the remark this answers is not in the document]
alfetopito: There ya go https://github.com/gnosis/gp-ui/pull/131

src/hooks/useErc20.ts
alfetopito: https://github.com/gnosis/gp-ui/pull/132

Keep in mind this PR depends on https://github.com/gnosis/gp-ui/pull/130

src/hooks/useErc20.ts
alfetopito: I was not aware of this. Looks fun!
Maybe talk about that tomorrow?
I have already way too many open waterfall PRs :sweat_smile: 
And this will probably clear out a bunch of code in those PRs... :shrug:

PR #129, opened 2021-02-08
66/refactor global state folders

alfetopito:
# Summary

Part of #66 
Follow up to https://github.com/gnosis/gp-ui/pull/115#pullrequestreview-584629547
Built on top of #117 

> This would imply to rename reducers-actions to state, and break the file into actions and reducer. I think we don't have updater, for now, and for v1 we wouldn't need to adapt it into that for now. But nice to have the concept in mind in case we want to do the same thing.

Only split into parts the new reducer (`erc20`).
Left the legacy reducers unchanged.
Will implement `hooks` next.


anxolin (commented):
Great!!

Just some minor comments

image


I think there's some duplication right? maybe u forgot to delete `erc20.ts`

Also, if `hooks.ts` is empty, I think we can delete it until/if we create hooks

Lastly, I think `types` can go in the `actions` imo

EDIT: I see `hooks.ts` is added in next PR, so forget that comment.

alfetopito:
> Great!!
> 
> Just some minor comments
> 
> image
> 
> I think there's some duplication right? maybe u forgot to delete `erc20.ts`

Yes, deleted in a follow up PR

> 
> Also, if `hooks.ts` is empty, I think we can delete it until/if we create hooks
> 
> Lastly, I think `types` can go in the `actions` imo

Ok, will update

> 
> EDIT: I see `hooks.ts` is added in next PR, so forget that comment. #130

Yep

anxolin (approved):
cool!

You can merge, I'm sure David will review on monday

src/state/erc20/reducer.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export function reducer(state: Erc20State, action: ReducerActionType): Erc20State {
+  switch (action.type) {
+    case 'SAVE_MULTIPLE_ERC20': {
+      // Clone current state
+      const map = new Map(state)
+
+      const { erc20s, networkId } = action.payload
+
+      erc20s.forEach((erc20) => {
```
W3stside, line 12: not sure if this logic should be in the reducer, it may be an anti-pattern and could be better in a `batchAction` where it's necessary. Thoughts @anxolin

W3stside (commented):
nice

src/state/erc20/reducer.ts
alfetopito: How would that look like?

anxolin: yes, I think is a fair concern, it also seemed like this reducer had some logic there, but I would argue this logic is a pure reducer. Ir receives the new data, and it returns the merged map with the new state

Or can u explain further your concern david?

W3stside: mm @alfetopito now I take it back, I had brushed over this and wasn't sure if this was pure or not but it is. I usually prefer `thin` reducers and `thicker` or fat actions / component logic. So in this case it would be removing the `saveMultipleErc20` action/reducer and looping + dispatching single actions inside a 
```ts
React.batchActions(() => {  
  erc20s.forEach(() => dispatch(singleErc20Action(...)))
})
```  

but i digress, that's too much rework and this is fine. just a thought tho 
@anxolin tagging u as well

alfetopito: Makes sense, but what would be the advantage?
In the end, the object will be recreated on every update, doesn't matter if updating 1 or more items, no?
If we batch, there's N state updates instead of 1 (ok, they are batched, but still)

I don't mind using the batched version, just want to understand why would that be better

PR #130, opened 2021-02-10
66/erc20 reducer hooks

alfetopito:
# Summary

Part of #66
Builds on top of/follow up to https://github.com/gnosis/gp-ui/pull/129

Added hooks for easier access to erc20 global state

Also added a hook for fetching multiple erc20s at the same time.


src/apps/explorer/components/OrderWidget/index.tsx
```diff
@@ -29,11 +29,16 @@ export const OrderWidget: React.FC = () => {
...
-  const { value: sellToken } = useErc20({ address: order?.sellToken, networkId: 4 })
-
-  console.log(`buy token`, buyToken)
-  console.log(`sell token`, sellToken)
+  // const { value: buyToken } = useErc20({ address: order?.buyToken, networkId: 4 })
+  // const { value: sellToken } = useErc20({ address: order?.sellToken, networkId: 4 })
+  // console.log(`buy token`, buyToken)
+  // console.log(`sell token`, sellToken)
+
+  const multiple = useMultipleErc20({
```
anxolin, line 37: this is also, just for testing the hook too, right? u will kill it after?

src/hooks/useErc20.ts
```diff
@@ -59,7 +65,92 @@ export function useErc20(params: Params): Return {
...
+ * Returns `error` with the error messages, if any.
+ */
+export function useMultipleErc20(params: UseMultipleErc20Params): Return<Record<string, SingleErc20State>> {
+  const { addresses, networkId } = params
+
+  const [isLoading, setIsLoading] = useState(false)
+  const [error, setError] = useState('')
+
+  const getMultipleErc20 = useGetMultipleErc20sFromGlobalState(networkId)
+  const saveErc20s = useSaveErc20sToGlobalState(networkId)
```
anxolin, line 93: mm, this is not how you get the state and change the state, right?
persisting a token should be just dispatching an action. Maybe I'm missing sth

src/hooks/useErc20.ts
```diff
@@ -59,7 +65,92 @@ export function useErc20(params: Params): Return {
...
+
+  useEffect(() => {
+    // only trigger network query if not yet running and there's anything not found yet
+    if (!running.current && toFetch.length > 0) {
+      loadFromNetwork(toFetch)
+    }
+  }, [loadFromNetwork, toFetch, saveErc20s])
+
+  // Not sure this `useMemo` is necessary
+  return useMemo(() => ({ isLoading, error, value: erc20s }), [erc20s, error, isLoading])
```
anxolin, line 155: this method is a big hard to read I think, can't be simplified?

src/hooks/useErc20.ts
```diff
@@ -59,7 +65,92 @@ export function useErc20(params: Params): Return {
...
+  )
+
+  useEffect(() => {
+    // only trigger network query if not yet running and there's anything not found yet
+    if (!running.current && toFetch.length > 0) {
+      loadFromNetwork(toFetch)
+    }
+  }, [loadFromNetwork, toFetch, saveErc20s])
+
+  // Not sure this `useMemo` is necessary
```
anxolin, line 154: no, I wouldn't add a useMemo, yes is more optimal but I thing nothing too big is being constructed here. Just a small function. 
So not worth it imo

src/hooks/useErc20.ts
```diff
@@ -59,7 +65,92 @@ export function useErc20(params: Params): Return {
...
+      setIsLoading(false)
+      running.current = false
+    },
+    [networkId, saveErc20s],
+  )
+
+  useEffect(() => {
+    // only trigger network query if not yet running and there's anything not found yet
+    if (!running.current && toFetch.length > 0) {
+      loadFromNetwork(toFetch)
```
anxolin, line 150: the function already has `toFetch` in it's clousure, so the param is not needed

src/hooks/useErc20.ts
```diff
@@ -59,7 +65,92 @@ export function useErc20(params: Params): Return {
...
+      const promises = toFetch.map(
+        async (address): Promise<SingleErc20State> => {
+          try {
+            const erc20 = await getErc20Info({ tokenAddress: address, networkId, web3, erc20Api })
+
+            return erc20
+          } catch (e) {
+            const msg = `Failed to fetch erc20 details for ${address} on network ${networkId}`
+            console.error(msg, e)
+            setError((errors) => [errors, msg].filter(Boolean).join('\n'))
```
anxolin, line 122: should `errors` parameter be in singular?, after all it's call setError

` [errors, msg].filter(Boolean).join('\n')` is a bit geeky?, maybe we errors should be an array of strings instead of a string. What if you want to later show it in a <ul> for example. Maybe it should even be an object with the address as a key. This way, if your token is null, you know you have an error message waiting for you

src/state/erc20/hooks.ts
```diff
@@ -0,0 +1,82 @@
...
+
+export type SingleErc20State = TokenErc20 | null
+
+/**
+ * Syntactic sugar to get erc20 from global state.
+ * Returns a function that fetches the TokenErc20 or null for the given address
+ *
+ * @param networkId The network id
+ */
+export const useGetErc20FromGlobalState = <State extends { erc20s: Erc20State }>(
```
anxolin, line 20: I think this method should be called `useErc20` and just get that specific erc20 from the global state. No not returning a callback so someone can call from outside

anxolin (commented):
The main concern is that the code feels a bit hard to read the flow, specially the hook for getting multiple tokens

src/state/erc20/hooks.ts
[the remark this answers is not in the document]
alfetopito: Makes sense. Just wanted to differentiate from simple accessing the global state from the hook that will try the global state and fetch from the network when not found.
Any name suggestions?

src/apps/explorer/components/OrderWidget/index.tsx
alfetopito: Yep, only for testing

src/hooks/useErc20.ts
alfetopito: Yes, that is. The hook `useSaveErc20sToGlobalState` returns a function that encapsulates the logic of dispatching the action.

Probably over engineered :sweat_smile: 

What do you suggest? Not have the hook and access the state directly?

src/hooks/useErc20.ts
alfetopito: I can maybe extract some inline functions outside. Will give it a try

src/hooks/useErc20.ts
alfetopito: Good call, will make it an object

src/state/erc20/hooks.ts
alfetopito: As I mentioned in a comment above, this name is to avoid clashing with the hook that will also fetch the token from the network if needed.

What should I call that one, then?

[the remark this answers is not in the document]
anxolin: `useErc20`?

if u have a collition in, let's say, the `hooks/useErc20`, u can always rename it just there in it's import to `useErc20State`

src/hooks/useErc20.ts
anxolin: If you want, we can leave it for another PR. 
Also, I don't mind to give it a try another time, maybe I convince myself after that that is not worth it :)

src/hooks/useErc20.ts
anxolin: haha, i saw my comment, and in the part I say 
` What if you want to later show it in a` there's a line break, cause I wrote an Html **UL** which was converted to HTML or sth messing with my comment

src/state/erc20/hooks.ts
anxolin: I close this one, to leave the other thread open

src/hooks/useErc20.ts
alfetopito: This hook is only to update the state `useSaveErc20sToGlobalState`. I renamed it to a shorted version - as well as the other hooks on this module. It's now `useSaveErc20s`.

I kept it returning a callback. The other hooks are now returning the object directly instead of the callback

src/state/erc20/hooks.ts
```diff
@@ -0,0 +1,79 @@
...
+export const useMultipleErc20s = <State extends { erc20s: Erc20State }>(
+  params: UseMultipleErc20Params,
+): Record<string, SingleErc20State> => {
+  const { addresses, networkId } = params
+  const [{ erc20s }] = useGlobalState<State>()
+
+  return useMemo(() => {
+    if (!networkId) {
+      // Return map of address => null
+      return addresses.reduce((acc, address) => ({ ...acc, [address]: null }), {})
```
anxolin, line 52: not opposed to returning this, but is this better than the empty object `{}`?

src/state/erc20/hooks.ts
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    return addresses.reduce(
+      (acc, address) => ({ ...acc, [address]: erc20s.get(buildErc20Key(networkId, address)) || null }),
```
anxolin, line 56: a bit hard to read, but not that bad. I personally find is ok more line breaks...

anxolin (approved):
Super good refactor! way nicer

PR #133, opened 2021-02-10
Network state

anxolin:
WIP:

- add networkId to the global state based on the URL
- adds convenient hooks and updater for the state
- Use the network id in the order

### DIFF
--- a/src/apps/explorer/ExplorerApp.tsx
+++ b/src/apps/explorer/ExplorerApp.tsx
@@ -11,7 +11,7 @@ import { rootReducer, INITIAL_STATE } from 'apps/explorer/state'
 import { GenericLayout } from 'components/layout'
 import { Navigation } from 'components/layout/GenericLayout/Navigation'
 import { Header } from 'components/layout/GenericLayout/Header'
-import { UpdateNetwork } from 'state/network'
+import { NetworkUpdater } from 'state/network'
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 const Router: typeof BrowserRouter & typeof HashRouter = (window as any).IS_IPFS ? HashRouter : BrowserRouter
@@ -59,7 +59,7 @@ const HEADER = (
 export const Updaters: React.FC = () => {
   return (
     <>
-      <UpdateNetwork />
+      <NetworkUpdater />
     </>
   )
 }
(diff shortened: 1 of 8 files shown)


src/state/network/updater.tsx
```diff
@@ -0,0 +1,40 @@
...
+    case 'rinkeby':
+      return Network.Rinkeby
+    case 'xdai':
+      return Network.xDAI
+    default:
+      return Network.Mainnet
+  }
+}
+
+export const NetworkUpdater: React.FC = () => {
```
alfetopito, line 21: Doesn't have children nor takes in any params. Use a simple function instead:

```suggestion
export function NetworkUpdater(): JSX.Element {
```

alfetopito: BTW, does it have to be a component? Can't we use it as a hook instead?

In my mind components are things that display something. Feels weird to have a component that'll always return `null`.

src/apps/explorer/ExplorerApp.tsx
```diff
@@ -55,18 +56,31 @@ const HEADER = (
   </Header>
 )
 
+export const Updaters: React.FC = () => {
+  return (
+    <>
+      <NetworkUpdater />
+    </>
```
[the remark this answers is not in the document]
anxolin: nup! good call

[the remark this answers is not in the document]
anxolin: it was cause I expect to add more updaters. We have kind of similar things but done in a different way. I like this pattern, to keep the updaters from the global state inside the folder of the state, and add them here.

I can delete while we don't add a second one..

src/state/network/updater.tsx
anxolin: Fair enough! I followed uniswap approach, but I think u are right, it feels more natural to use a hook. Addressed

src/apps/explorer/state.ts
```diff
@@ -1,17 +1,23 @@
...
 
 export type ExplorerAppState = GlobalState & {
   erc20s: Erc20State
+  networkId: Network | null
 }
 
 export const INITIAL_STATE = (): ExplorerAppState => ({
   ...GLOBAL_INITIAL_STATE(),
   erc20s: INITIAL_ERC20_STATE,
+  networkId: null,
```
alfetopito, line 16: Why not start with state == mainnet?

alfetopito (approved):
Good to go, only 1 minor comment

src/apps/explorer/state.ts
anxolin: Oh, sorry I read this after merging!
Good question!

I don't think to trigger calls or anything heavy, or re-rendering before a tiny and fast app warm up.
So it's null for a few milliseconds, then it should have a networkId

alfetopito: It's alright for now. We can revisit later.

W3stside:
@anxolin redirect `/mainnet/` in URL to default mainnet url scheme

PR #136, opened 2021-02-11
Quick and dirty order details page

alfetopito:
# Summary

Wiring up order details in a quick and dirty way, to be ready for trading challenge tomorrow (2021-02-11)

Will clean this up properly later

## Open order
screenshot_2021-02-11_10-49-53

## Filled order
screenshot_2021-02-11_10-50-51

PR #137, opened 2021-02-11
Web3 instance per network

alfetopito:
# Summary

Builds on top of https://github.com/gnosis/gp-ui/pull/136 (:sweat_drops: :fallen_leaf:)

:warning: merge only after #136 

---

Updating web3 instance provider when network changes.

This PR adds 2 approaches:

1. Create a web3 instance per network
2. Use a single web3 instance, and update its provider when the network changes

Currently approach `2` is in use, although both were added.
Check their implementations for more details, respectively https://github.com/gnosis/gp-ui/blob/9570c86e3f09d3e78fd456644ed9b75f355afe68/src/api/web3/index.ts and https://github.com/gnosis/gp-ui/blob/9570c86e3f09d3e78fd456644ed9b75f355afe68/src/api/web3/hooks.ts

# Testing

Compare the same order on the:
- Previous PR: https://pr136--gpui.review.gnosisdev.com/rinkeby/orders/0x3960fcd923574dcd0c0f40f9bd20485a2c8b4f85128eedb3c3d53960691d1d7d5b0abe214ab7875562adee331deff0fe1912fe4260257f55/
screenshot_2021-02-11_13-47-00

- Current PR: https://pr137--gpui.review.gnosisdev.com/rinkeby/orders/0x3960fcd923574dcd0c0f40f9bd20485a2c8b4f85128eedb3c3d53960691d1d7d5b0abe214ab7875562adee331deff0fe1912fe4260257f55/
screenshot_2021-02-11_13-47-07

### DIFF
--- a/src/api/web3/hooks.ts
+++ b/src/api/web3/hooks.ts
@@ -1,39 +1,16 @@
 import { useMemo } from 'react'
 import Web3 from 'web3'
 
-import { getNetworkFromId } from '@gnosis.pm/dex-js'
-
 import { useNetworkId } from 'state/network'
 
-import { Network } from 'types'
+import { createWeb3Api, getProviderByNetwork } from '.'
 
-import { INFURA_ID } from 'const'
-
-import { createWeb3Api } from '.'
-
-function infuraProvider(networkId: Network): string {
-  // INFURA_ID relies on mesa `config` file logic.
-  // We can be independent of that config by relying on the env var directly
-  if (!INFURA_ID) {
-    throw new Error(`INFURA_ID not set`)
-  }
-  return `wss://${getNetworkFromId(networkId).toLowerCase()}.infura.io/ws/v3/${INFURA_ID}`
-}
-
-function infuraProviderByNetworkId(networkId: Network | null): string | undefined {
-  switch (networkId) {
-    case Network.Mainnet:
-    case Network.Rinkeby:
-    case Network.xDAI:
-      return infuraProvider(networkId)
-    default:
-      return undefined
-  }
-}
-
+// Approach 1: create one web3 instance per network
+// Must be used inside components, since it's a hook
+// Can also be used outside by calling `createWeb3Api` directly
 export function useWeb3(): Web3 {
   const networkId = useNetworkId()
-  const provider = infuraProviderByNetworkId(networkId)
+  const provider = getProviderByNetwork(networkId)
 
   // createWeb3 caches instances per provider, safe to call multiple times
   return useMemo(() => createWeb3Api(provider), [provider])
--- a/src/api/web3/hooks.ts
+++ b/src/api/web3/hooks.ts
@@ -1,0 +1,40 @@
+import { useMemo } from 'react'
+import Web3 from 'web3'
+
+import { getNetworkFromId } from '@gnosis.pm/dex-js'
+
+import { useNetworkId } from 'state/network'
+
+import { Network } from 'types'
+
+import { INFURA_ID } from 'const'
+
+import { createWeb3Api } from '.'
+
+function infuraProvider(networkId: Network): string {
+  // INFURA_ID relies on mesa `config` file logic.
+  // We can be independent of that config by relying on the env var directly
+  if (!INFURA_ID) {
+    throw new Error(`INFURA_ID not set`)
+  }
+  return `wss://${getNetworkFromId(networkId).toLowerCase()}.infura.io/ws/v3/${INFURA_ID}`
+}
+
+function infuraProviderByNetworkId(networkId: Network | null): string | undefined {
+  switch (networkId) {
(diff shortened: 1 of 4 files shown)


src/api/web3/hooks.ts
```diff
@@ -0,0 +1,17 @@
+import { useMemo } from 'react'
```
anxolin, line 1: Are we also adding hooks in the API? 

I thought we were directly adding those to `hooks/`, maybe you are right and it makes sense and is coherent with what we do in the state. 

I think we should at some point review our current hooks and re-think about a structure for them.
Maybe some are even useful only in the context of v1, so they should be moved to `gp-v1` app probably

src/api/web3/index.ts
```diff
@@ -1,13 +1,23 @@
...
-import { ETH_NODE_URL } from 'const'
+import { Network } from 'types'
+
+import { ETH_NODE_URL, INFURA_ID } from 'const'
 
 // TODO connect to mainnet if we need AUTOCONNECT at all
 export const getDefaultProvider = (): string | null => (process.env.NODE_ENV === 'test' ? null : ETH_NODE_URL)
 
-export function createWeb3Api(): Web3 {
+const web3cache = {}
```
anxolin, line 11: Probably you would want to be more specific about the type, right?
Like using `Record<string, Web3>`

src/api/web3/index.ts
```diff
@@ -16,5 +26,42 @@ export function createWeb3Api(): Web3 {
...
+  // INFURA_ID relies on mesa `config` file logic.
+  // We can be independent of that config by relying on the env var directly
+  if (!INFURA_ID) {
+    throw new Error(`INFURA_ID not set`)
+  }
+  return `wss://${getNetworkFromId(networkId).toLowerCase()}.infura.io/ws/v3/${INFURA_ID}`
+}
+
+// For now only infura provider is available
+export function getProviderByNetwork(networkId: Network | null): string | undefined {
```
anxolin, line 44: Me like!

src/state/network/updater.tsx
```diff
@@ -33,6 +35,7 @@ export const NetworkUpdater: React.FC = () => {
     // Update the network if it's different
     if (currentNetworkId !== networkId) {
       dispatch(setNetwork(networkId))
+      updateWeb3Provider(web3, networkId)
```
anxolin, line 38: mmm, it works, it's even kind of nice, but I think it has a problem.

This state `network` is super specific, probably I should have called that `network-from-path`
It's an opt-in functionality apps can just add in their combine reducer (+add the updater) and benefit from it. 

You are assuming here all users of this functionality will use a web3 wallet.
- What if don't have wallet, and I don't want one
- What if we want to use EtherJS :) 
- What our API of web3 gets deprecated in favour of another one better, but we don't want to adapt all the apps.

Another approach is to add an updater for the API, this updater is simple. It watches the global state for change.
Then, if u need just the `network-from-path` you just use that, if you need the wallet to, you add also the updater for that.

anxolin (approved):
Great PR, I like it very much. I added some comments and I think I'll address some of them in a PR since I can't wait for mefing this one 🚀

src/api/web3/hooks.ts
alfetopito: Yeah, wasn't totally sure about this one either.
One the bright side, this is the approach 1, which I'm starting to like less in favour of approach 2.

In any case, we can always move this to hooks/ instead of leaving it in the API.

src/api/web3/index.ts
alfetopito: This is only needed if we go for approach 1 (1 instance per network).
We can do that anyway, but I'm more in favor of going for approach 2

src/state/network/updater.tsx
alfetopito: > You are assuming here all users of this functionality will use a web3 wallet.

Well, not really. This change is precisely for when the users don't have a connected wallet, and the app needs to change the network based on the currently selected network.

I expect that when users connect a wallet, it'll provide us with the appropriated provider. 
Although, we are not there yet, so not really thinking about this =P

> Another approach is to add an updater for the API, this updater is simple. It watches the global state for change.
Then, if u need just the network-from-path you just use that, if you need the wallet to, you add also the updater for that.

So you suggest a global state for `web3`, that depends on the `network` state?

---

I'm not gonna defend this as the perfect solution, as I'm not too happy with this either.
Happy to do whatever makes more sense.

anxolin: My point is that here you assume there's a Web3Provider provider.

My network state was isolated, not related to web3provider in any way. So now you assume all the apps need to call `updateWeb3Provider` when the `networkId` changes

For me, this is a side effect on the network changing, I'll probably make a more specific suggestion. So forget for now

PR #141, opened 2021-02-12
Style updates to the GP home.

biocom:
Updated some elements on the homepage:
- Higher res home image
- Allow for normal scrolling (instead of fixed container)

Screen Shot 2021-02-12 at 12 36 31


anxolin (approved):
Thanks, merging so it's in the demo

PR #145, opened 2021-02-12
92/refactor components structure

alfetopito:
# Summary

Part of #92 

Refactoring components structure.

Lot's of files changed, I know.
But I guarantee, there's no code change introduced. Only moving components around.

Issue #92, opened 2021-01-29
Component name convention


W3stside:
From @anxolin

I vote against this naming convention. 
I prefer to call the dumb components without any suffix, they are just normal components.

First naming it `OrderWidgetView` u imply there's a 1<>1 relationship between OrderWidget and OrderWidgetView, which might be true here at least for now, although is not sth we always want to do. 

Another concern would be that the names get unnecesarilly complex. `MyCoolComponentWidgetViewStyledWrapperHoH` (disclaimer: some names in this film may or may not have been dramatized)

_Originally posted by @anxolin in https://github.com/gnosis/gp-ui/pull/91#discussion_r566684624_

Task: reconsider nomenclature

alfetopito:
So, what are your suggestions?

W3stside:
I am of no real opinion here, i do agree with @anxolin that a decoupled generic dumb component shouldbt be named too specifically. so for the OrdersWidgetView, if its always meant to be coupled to that then keep it else just name it DataView or sth

alfetopito:
My goal with this decoupling is simply to isolate business logic from presentation and make testing (namely, Storybook) easier.
It's not intended to be generic enough to be used in isolation, thus there's still a high level of coupling between them.

As for the name, I don't care, whatever you prefer.

I can suggest a shorter version, that indicates what it is, without decoupling from the parent component:
From `OrderWidgetView` to `View` (which you probably won't like :shrug: )

anxolin:
Decoupling I think is a good idea. I think is nice to have a component that represents the whole order. So I agree on that.
For me they are 100% independent, at least the dumb is independent of the Smart. The smart of course depend on the dumb.

For the naming, I would just eliminate any "sufix", just call it what you think it represents.

`OrderDetails` sound good to me. It would receive the order data, the fills array, and some callback for some actions that u can do in that page (if we can do any action) --> i.e. if we paginate the fills, which we can assume we don't do for now.

alfetopito:
> `OrderDetails` sound good to me

Well, that's the name of the subcomponent https://github.com/gnosis/gp-ui/blob/develop/src/apps/explorer/components/OrderWidget/OrderDetails.tsx which has only the order table.
The fills are supposed to be a different sub-component, let's say `OrderFills`.

Which together compose `OrderWidget`

anxolin:
[...]

PR #146, opened 2021-02-12
92/refactor order widget structure

alfetopito:
# Summary

Part of #92 
Depends on #145 (:sweat_drops: :fallen_leaf:)

Moved OrderDetails to `components/orders` and started refactoring it.

New sub-component `OrderPriceDisplay`, with associated storybook stories.
screenshot_2021-02-12_14-56-56

# Testing

- Check the storybook https://pr146--gpui.review.gnosisdev.com/storybook/?path=/docs/orders-orderpricedisplay--default
- Check the live app https://pr146--gpui.review.gnosisdev.com/rinkeby/orders/0xa6223509c0a41f992995d8781c3452ae8e2f55640b21d057fc02c7fb5a1674b35b0abe214ab7875562adee331deff0fe1912fe4260246928/

Switch button fully self-contained.
Inverting the price does not cause OrderDetails re-render.


src/components/orders/OrderPriceDisplay/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+export type Props = {
+  type?: 'limit' | 'executed'
+  order: RawOrder
+  buyToken: TokenErc20
+  sellToken: TokenErc20
+  invertedPrice?: boolean
+}
+
+export function OrderPriceDisplay(props: Props): JSX.Element {
+  const { type = 'limit', order, buyToken, sellToken, invertedPrice: initialInvertedPrice } = props
```
W3stside, line 34:
```suggestion
  const { type = 'limit', order, buyToken, sellToken, invertedPrice: initialInvertedPrice = false } = props
```
Does that work?

src/components/orders/OrderPriceDisplay/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  const [invertedPrice, setInvertedPrice] = useState(initialInvertedPrice || false)
```
W3stside, line 36: Assuming my suggestion on `L34`

```suggestion
  const [invertedPrice, setInvertedPrice] = useState(initialInvertedPrice)
```

src/components/orders/OrderPriceDisplay/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const invertPrice = useCallback(() => setInvertedPrice((curr) => !curr), [])
```
W3stside, line 37: nit: probably not required to useCallback on this

https://kentcdodds.com/blog/usememo-and-usecallback

src/components/orders/OrderPriceDisplay/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+    const price = getPrice({
+      order,
+      buyTokenDecimals: buyToken.decimals,
+      sellTokenDecimals: sellToken.decimals,
+      inverted: invertedPrice,
+    })
+    return formatSmart({ amount: price.toString(), precision: 0, smallLimit: '0.000001', decimals: 7 })
+  }, [buyToken.decimals, getPrice, invertedPrice, order, sellToken.decimals])
+
+  const [baseSymbol, quoteSymbol] = useMemo(() => {
```
W3stside, line 51: maybe some of this stuff could be it's own hook?

W3stside (approved):

src/components/orders/OrderPriceDisplay/index.tsx
```diff
@@ -0,0 +1,63 @@
+import React, { useCallback, useMemo, useState } from 'react'
+import styled from 'styled-components'
+import { faExchangeAlt } from '@fortawesome/free-solid-svg-icons'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+
+import { formatSmart, safeTokenName, TokenErc20 } from '@gnosis.pm/dex-js'
+
+import { getOrderExecutedPrice, getOrderLimitPrice, RawOrder } from 'api/operator'
```
anxolin, line 8: I still think it's strange that these utils are in the API. Being used in the component should be reason enough to move them to the utils. 

The 1st thing I thought was, that the component was getting the prices from the API, so I was concerned. But those are static functions...

Alternatively, if those prices are used everywhere, maybe we should have them as part of the data

src/components/orders/OrderPriceDisplay/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+const Icon = styled(FontAwesomeIcon)`
+  background: ${({ theme }): string => theme.textSecondary2}33; /* 33==20% transparency in hex */
+  border-radius: 1rem;
+  width: 2rem !important; /* FontAwesome sets it to 1em with higher specificity */
+  height: 2rem;
+  padding: 0.4rem;
+  margin-left: 0.5rem;
+`
+
+export type Props = {
```
anxolin, line 25: Are we sure we want this signature for the proos? It feels weird to hand in an order and a type. 
why not to hand in a fraction: numerator, denominator and allow the component to display any prices (not just order prices)

Will this work with fills? they also have execution prices? it feels a bit rigid. Although maybe for this ones, it will be another component, because the invert icon is not in it... (or maybe it is and is just to not show it, not sure)

src/components/orders/OrderPriceDisplay/index.tsx
alfetopito: It does!

src/components/orders/OrderPriceDisplay/index.tsx
alfetopito: Sooo, I gather from the article that pretty much all the memoizations in this component are unnecessary?

src/components/orders/OrderPriceDisplay/index.tsx
alfetopito: As I said before, those utils are intended to be what in an OO API are object methods.

> Alternatively, if those prices are used everywhere, maybe we should have them as part of the data

The initial approach was to work with the Raw API order data, but I have nothing against adding those calculated values to a `InternalOrder` type, or whatever we call it.

src/components/orders/OrderPriceDisplay/index.tsx
alfetopito: I agree with your concerns.

I ended up with this mainly for 2 reasons:
1. To encapsulate the price inversion inside the component, to avoid parent re-renders
2. I already had implemented the utils fns that take a parameter to invert the price, thus the `type: 'limit' | 'executed'`

So, the implementation got constrained by those decisions.

I'll rewrite this be more generic, and invert the price without relying on the nice (but useless) invert price param utils fn:

```typescript
type Props = {
  buyAmount: string
  buyToken: TokenErc20
  sellAmount: string
  sellToken: TokenErc20
  priceInverted?: boolean  // in case we want to control from the outside, like on the fills
  showInvertIcon?: boolean  // in case we don't want to show the invert button, like on the fills
}
```

src/components/orders/OrderPriceDisplay/index.tsx
alfetopito: I don't think that's necessary now, but let me know whether it's still needed after the refactor.

src/components/orders/OrderPriceDisplay/index.tsx
alfetopito: I'll send out a different PR for addressing this part

alfetopito: https://github.com/gnosis/gp-ui/pull/161

src/components/orders/OrderPriceDisplay/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+  buyAmount: string | BigNumber
+  buyToken: TokenErc20
+  sellAmount: string | BigNumber
+  sellToken: TokenErc20
+  isPriceInverted?: boolean
+  withoutInvertButton?: boolean
```
anxolin, line 30: I would avoid "negation" names. They can make code harder to read, specially when we start to have also complex boolean expressions

But even with a simple true/false value is harder to read:
```bash
withoutInvertButton={false}

// As opposed to
showInvertButton={true}

// ..or
invertButtonVisible={true}
```

anxolin (commented):
APPROVED: with minor comment.

* I won't specify every time, but this is a TEXTUAL APPROVE, so mergify don't do it's thing and leaves it up to Leandro if he addresses the comment before merging

src/components/orders/OrderPriceDisplay/index.tsx
W3stside: up to you i think the margins here are tiny. but :shrug:

PR #158, opened 2021-02-15
Network state url

anxolin:
## 🚨 don't merge

Does anyone know what's going on here?

-----

So this was an attempt to try to decouple how we select the current network, and how we setup the web3 instance.

For this, I added web3 to the global state. This way, different Updaters can implement different logics.
For the explorer I did this Updater that will watch the `networkId` of `state/network-in-url` and update the web3 using it's value

Other apps could have other logics on setting the web3 state. 

Problem I faced and I cannot see right now is, that after my changes I see 2 super strange behaviours:
- The app is being re-rendered in an infinity loop! This happens because the `withGlobalContext` is re-rendered. I would think that someone mutates the state each time so it detects a change and re-renders ??? (no sure)
- Other weird thing is what I show in this image, somehow some state is disappearing. I though this was not possible if our reducers only act in their piece of state:
image


W3stside:
Can probably re enable the default middlewares and debug this. Likely some Updater somewhere is either mutating state or sending unserialisable stuff to state causing rerenders

src/state/web3/updater.tsx
```diff
@@ -0,0 +1,58 @@
...
+  }, [currentNetworkId, networkId, dispatch, web3])
+
+  return null
+}
+
+function updateWeb3(networkId: Network | null, web3: Web3 | null, dispatch: React.Dispatch<ReducerActionType>): void {
+  let newWeb3: Web3 | null | undefined
+  if (!networkId) {
+    // If there's no network, we set web3 to null
+    newWeb3 = null
```
alfetopito, line 36: Why null? Can't we default to mainnet?

src/state/web3/updater.tsx
```diff
@@ -0,0 +1,58 @@
...
+      web3.setProvider(provider)
+    } else {
+      // Unsupported network
+      newWeb3 = null
+    }
+  }
+
+  // Update web3
+  if (newWeb3 !== undefined) {
+    dispatch(setWeb3(newWeb3))
```
alfetopito, line 56: So, you do want to set web3 to `null` but not `undefined`?

W3stside:
for whomever has capactiy to debug this, try this: check `custom/state/index` and check that the `defaultMiddleware` thing has all middlewares (except thiunk) enabled, sth like `serialise` middleware and `mutation` middleware, then check the stack trace for errors. I THINK an `updater.tsx` somewhere is adding non-serialisable state to redux which is causeing the whole shabang to re-render blah blah @alfetopito @anxolin

PR #161, opened 2021-02-17
92/use order and erc20s hook

alfetopito:
# Summary

Part of #92
Waterfalls into https://github.com/gnosis/gp-ui/pull/146/files
Follow up to https://github.com/gnosis/gp-ui/pull/146#discussion_r576417353

New order type, transforming between `RawOrder` and `Order` (the new type) and new hook vastly improving `OrderWidget`.

Still more refactor to be done. We'll get there.

### DIFF
--- a/src/api/operator/types.ts
+++ b/src/api/operator/types.ts
@@ -57,7 +57,7 @@ export type Order = Pick<RawOrder, 'owner' | 'uid' | 'appData' | 'kind' | 'parti
   feeAmount: BigNumber
   executedFeeAmount: BigNumber
   cancelled: boolean
-  // status: OrderStatus // better not have the status because it depends on `expirationDate`
+  status: OrderStatus
   filledAmount: BigNumber
   filledPercentage: BigNumber
   // limitPrice: BigNumber  // cant have prices because they depend on token decimals
--- a/src/api/operator/types.ts
+++ b/src/api/operator/types.ts
@@ -1,3 +1,5 @@
+import BigNumber from 'bignumber.js'
+
 import { Network } from 'types'
 
 export type OrderID = string
@@ -36,6 +38,30 @@ export type RawOrder = {
   kind: OrderKind
   partiallyFillable: boolean
   signature: string
+}
+
+/**
+ * Enriched Order type.
+ * Applies some transformations on the raw api data.
+ * Some fields are kept as is.
+ */
+export type Order = Pick<RawOrder, 'owner' | 'uid' | 'appData' | 'kind' | 'partiallyFillable' | 'signature'> & {
+  creationDate: Date
+  expirationDate: Date
+  buyTokenAddress: string
+  sellTokenAddress: string
+  buyAmount: BigNumber
+  sellAmount: BigNumber
+  executedBuyAmount: BigNumber
+  executedSellAmount: BigNumber
+  feeAmount: BigNumber
+  executedFeeAmount: BigNumber
+  cancelled: boolean
+  // status: OrderStatus // better not have the status because it depends on `expirationDate`
+  filledAmount: BigNumber
+  filledPercentage: BigNumber
+  // limitPrice: BigNumber  // cant have prices because they depend on token decimals
+  // executedPrice: BigNumber
 }
 
 type WithNetworkId = { networkId: Network }
--- a/src/api/operator/types.ts
+++ b/src/api/operator/types.ts
@@ -60,8 +60,6 @@ export type Order = Pick<RawOrder, 'owner' | 'uid' | 'appData' | 'kind' | 'parti
   status: OrderStatus
   filledAmount: BigNumber
   filledPercentage: BigNumber
-  // limitPrice: BigNumber  // cant have prices because they depend on token decimals
-  // executedPrice: BigNumber
 }
 
 type WithNetworkId = { networkId: Network }
--- a/src/api/operator/types.ts
+++ b/src/api/operator/types.ts
@@ -1,5 +1,7 @@
 import BigNumber from 'bignumber.js'
 
+import { TokenErc20 } from '@gnosis.pm/dex-js'
+
 import { Network } from 'types'
 
 export type OrderID = string
@@ -49,7 +51,9 @@ export type Order = Pick<RawOrder, 'owner' | 'uid' | 'appData' | 'kind' | 'parti
   creationDate: Date
   expirationDate: Date
   buyTokenAddress: string
+  buyToken?: TokenErc20 | null // undefined when not set, null when not found
   sellTokenAddress: string
+  sellToken?: TokenErc20 | null
   buyAmount: BigNumber
   sellAmount: BigNumber
   executedBuyAmount: BigNumber
--- a/src/apps/explorer/components/OrderWidget/index.tsx
+++ b/src/apps/explorer/components/OrderWidget/index.tsx
@@ -2,12 +2,15 @@ import React from 'react'
 import { useParams } from 'react-router'
 
 import { useOrderAndErc20s } from 'hooks/useOperatorOrder'
+
+import { ORDER_QUERY_INTERVAL } from 'apps/explorer/const'
+
 import { OrderWidgetView } from './view'
 
 export const OrderWidget: React.FC = () => {
   const { orderId } = useParams<{ orderId: string }>()
 
-  const { order, buyToken, sellToken, isLoading, error } = useOrderAndErc20s(orderId)
+  const { order, buyToken, sellToken, isLoading, error } = useOrderAndErc20s(orderId, ORDER_QUERY_INTERVAL)
 
   return (
     <OrderWidgetView
--- a/src/apps/explorer/components/OrderWidget/index.tsx
+++ b/src/apps/explorer/components/OrderWidget/index.tsx
@@ -10,15 +10,7 @@ import { OrderWidgetView } from './view'
 export const OrderWidget: React.FC = () => {
   const { orderId } = useParams<{ orderId: string }>()
 
-  const { order, buyToken, sellToken, isLoading, error } = useOrderAndErc20s(orderId, ORDER_QUERY_INTERVAL)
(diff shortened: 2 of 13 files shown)


src/api/operator/types.ts
```diff
@@ -38,6 +40,30 @@ export type RawOrder = {
...
+  buyTokenAddress: string
+  sellTokenAddress: string
+  buyAmount: BigNumber
+  sellAmount: BigNumber
+  executedBuyAmount: BigNumber
+  executedSellAmount: BigNumber
+  feeAmount: BigNumber
+  executedFeeAmount: BigNumber
+  cancelled: boolean
+  // status: OrderStatus // better not have the status because it depends on `expirationDate`
```
W3stside, line 60: are you keeping these commented props for a reason

alfetopito: So we can talk about it :)
Will remove

src/api/operator/types.ts
```diff
@@ -38,6 +40,28 @@ export type RawOrder = {
   signature: string
 }
 
+/**
+ * Enriched Order type.
+ * Applies some transformations on the raw api data.
+ * Some fields are kept as is.
+ */
+export type Order = Pick<RawOrder, 'owner' | 'uid' | 'appData' | 'kind' | 'partiallyFillable' | 'signature'> & {
```
anxolin, line 48: 🍝😋 delicious

src/api/operator/types.ts
```diff
@@ -38,6 +40,28 @@ export type RawOrder = {
...
+  creationDate: Date
+  expirationDate: Date
+  buyTokenAddress: string
+  sellTokenAddress: string
```
anxolin, line 52: if we enrich this, should we also enrich the token? not sure, just thinking...

src/apps/explorer/components/OrderWidget/index.tsx
```diff
@@ -1,62 +1,24 @@
...
-    addresses: [order?.buyToken || '', order?.sellToken || ''].filter(Boolean),
-    networkId: networkIdOrDefault,
-  })
-  const [buyToken, sellToken] = useMemo(() => {
-    const buyToken = erc20s[order?.buyToken || ''] || undefined
-    const sellToken = erc20s[order?.sellToken || ''] || undefined
-
-    return [buyToken, sellToken]
-  }, [order?.buyToken, order?.sellToken, erc20s])
+  const { order, buyToken, sellToken, isLoading, error } = useOrderAndErc20s(orderId, ORDER_QUERY_INTERVAL)
```
anxolin, line 13: it feels right when you can kill this much logic from the component

anxolin, line 13: Related to my comment above. This hooks looks fairly easy to use. Although one could think that buyToken, sellToken are part of the order. So ideally, the use would be:
```
const { order, isLoading, error } = useOrderAndErc20s(orderId, ORDER_QUERY_INTERVAL)

order.sellTokenAddress // address, we can leave it or kill this prop since it's derived from the sellToken
order.sellToken // token
```

src/components/orders/OrderDetails/index.tsx
```diff
@@ -91,11 +99,11 @@ export function OrderDetails(props: Props): JSX.Element {
           </tr>
           <tr>
             <td>Sell amount</td>
-            <td>{`${formatSmart(sellAmount, sellToken.decimals)} ${sellToken.symbol}`}</td>
+            <td>{`${formatSmart(sellAmount.toString(), sellToken.decimals)} ${sellToken.symbol}`}</td>
```
anxolin, line 102: ubernit. not important cause it's always hard to make changes to dex-js, and this one is too little. But if we touch it any time soon, maybe we want BigInt support (to save us the toString every time we format ---> no deal at all)

src/hooks/useOperatorOrder.ts
```diff
@@ -0,0 +1,99 @@
...
+  isLoading: boolean
+  forceUpdate: () => void
+}
+
+export function useOrder(orderId: string): UseOrderResult {
+  const [isLoading, setIsLoading] = useState(false)
+  const [error, setError] = useState('')
+  const [order, setOrder] = useState<Order | null>(null)
+  // Hack to force component to update itself on demand
+  const [forcedUpdate, setForcedUpdate] = useState({})
```
anxolin, line 23: I don't see where you use this

src/hooks/useOperatorOrder.ts
```diff
@@ -0,0 +1,99 @@
...
+ * @param orderId The order id
+ * @param updateInterval [Optional] How often should try to update the order
+ */
+export function useOrderAndErc20s(orderId: string, updateInterval = 0): UseOrderAndErc20sResult {
+  const networkId = useNetworkId() || undefined
+
+  const { order, isLoading: isOrderLoading, error: orderError, forceUpdate } = useOrder(orderId)
+  const addresses = order ? [order.buyTokenAddress, order.sellTokenAddress] : []
+
+  // TODO: consume errors obj
```
anxolin, line 76: u leave this one for another PR?

Maybe is just exposing it directly --> return erc20Errors also

src/hooks/useOperatorOrder.ts
anxolin: EDiT: All right, I see you call this from the other useOrder...

src/hooks/useOperatorOrder.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  useEffect(() => {
+    let intervalId: NodeJS.Timeout | null = null
+
+    // Only start the interval when:
+    // 1. Hook is configured to do so (`updateInterval` > 0)
+    // 2. Order exists
+    // 3. Order is not expired
+    if (updateInterval && order && order.expirationDate.getTime() > Date.now()) {
+      intervalId = setInterval(forceUpdate, updateInterval)
```
anxolin, line 90: Shouldn't this logic be on useOrder already? so you pass that hook the `updateInterval`

This way, you have the 4 options of use:
- basic order --> interval = 0
- basic order --> interval = 20s (for example)
- detailed order --> interval = 0
- detailed order --> interval = 20s (for example)

anxolin (commented):
Great work! mostly approved, just some discussions, but I'm happy with the current approach

src/api/operator/types.ts
alfetopito: I though about this.
The reasons for keeping it out of the `Order` are:
1. Token details are not part of the raw order, thus will have to be added later
2. Avoid nested objects so React is better at identifying obj updates/component redraw

For `1`, it's easy to address, since I'm fetching the token details in the same hook
For `2`, haven't really measured the performance impact. Was more of something I'm aware but don't fully know how it works.

None are strong reasons though. Will update.

src/apps/explorer/components/OrderWidget/index.tsx
alfetopito: It was all temporary throw away code anyway :)

src/components/orders/OrderDetails/index.tsx
alfetopito: Agreed. https://github.com/gnosis/gp-ui/issues/179

src/hooks/useOperatorOrder.ts
alfetopito: Yeah, should do something about it.
The issue is that this error is a mapping of address -> error msg

Think the best thing to do is to merge the order error with this obj and return the obj instead of a string.

src/hooks/useOperatorOrder.ts
alfetopito: Right, makes sense to move the update logic there.
I just don't get your 4 options. What do you mean?

src/api/operator/types.ts
alfetopito: :pie:

src/hooks/useOperatorOrder.ts
alfetopito: Updated and still don't get your options :D

src/api/operator/types.ts
anxolin: It was ok too. I didn't have strong opinion on this one either. I see reasons to do it and not do it

src/hooks/useOperatorOrder.ts
anxolin: ahaha, does't matter :) 

I was just showing that if you move it to the basic order hook, then anyone can use it as it is today, just pass a 0 to the interval. And then allow to set an interval. So it has both uses.

And the signature of the detailed one doesn't change. So it's win win.

anxolin (approved):

src/apps/explorer/components/OrderWidget/view.tsx
```diff
@@ -14,27 +12,26 @@ const Wrapper = styled.div`
...
     <Wrapper>
       <h2>Order details</h2>
       {/* TODO: create common loading indicator */}
-      {isLoading && <FontAwesomeIcon icon={faSpinner} spin size="5x" />}
-      {order && !isLoading && buyToken && sellToken && (
-        <OrderDetails order={order} buyToken={buyToken} sellToken={sellToken} />
-      )}
+      {order?.buyToken && order?.sellToken && <OrderDetails order={order} />}
       {!order && !isLoading && <p>Order not found</p>}
+      {!isLoading && order && (!order?.buyToken || !order?.sellToken) && <p>Not able to load tokens</p>}
```
W3stside, line 29: don't think you need `order` here if you use the `nullish coalesence` of `!order?.buToken` and sell

W3stside (commented):
approve

src/apps/explorer/components/OrderWidget/view.tsx
alfetopito: I actually need it. Only want to show it when one of the tokens failed loading.
Otherwise when the order is not found both lines 28 and 29 will be rendered.

PR #181, opened 2021-02-19
Misc order details changes

alfetopito:
# Summary

Waterfalls into https://github.com/gnosis/gp-ui/pull/162

Misc changes to details page I was already working on and are too small to deserve an issue on their own:
- Updated Type row
screenshot_2021-02-18_16-57-03
- Displaying a short id instead of the full uid
screenshot_2021-02-18_16-56-56
- Added Execution price row to fill or kill orders (`-` when not filled)
screenshot_2021-02-18_16-57-10

### DIFF
--- a/src/utils/format.ts
+++ b/src/utils/format.ts
@@ -200,3 +200,10 @@ export function formatPriceWithFloor(price: BigNumber): string {
   const displayPrice = amountToPrecisionDown(price, DEFAULT_DECIMALS).toString(10)
   return price.gt(LOW_PRICE_FLOOR) ? displayPrice : '< ' + LOW_PRICE_FLOOR.toString(10)
 }
+
+export function capitalize(sentence: string): string {
+  return sentence
+    .split(' ')
+    .map((word) => word[0].toUpperCase() + word.slice(1).toLowerCase())
+    .join(' ')
+}
--- a/test/data/operator.ts
+++ b/test/data/operator.ts
@@ -29,6 +29,7 @@ export const RAW_ORDER: RawOrder = {
 
 export const RICH_ORDER: Order = {
   ...RAW_ORDER,
+  shortId: 'adef89ade',
   creationDate: new Date(RAW_ORDER.creationDate),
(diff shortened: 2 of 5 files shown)


test/data/operator.ts
```diff
@@ -29,6 +29,7 @@ export const RAW_ORDER: RawOrder = {
 
 export const RICH_ORDER: Order = {
   ...RAW_ORDER,
+  shortId: 'adef89ade',
```
anxolin, line 32: cheatingggg

src/utils/format.ts
```diff
@@ -200,3 +200,10 @@ export function formatPriceWithFloor(price: BigNumber): string {
   const displayPrice = amountToPrecisionDown(price, DEFAULT_DECIMALS).toString(10)
   return price.gt(LOW_PRICE_FLOOR) ? displayPrice : '< ' + LOW_PRICE_FLOOR.toString(10)
 }
+
+export function capitalize(sentence: string): string {
+  return sentence
+    .split(' ')
+    .map((word) => word[0].toUpperCase() + word.slice(1).toLowerCase())
```
anxolin, line 207: I must be reading it wrong, cause reading this code I would expect to be Sell Order, not "Sell order"

what am i missing?
image

anxolin (approved):
Approved

minor one: pls use the order id to generate the short order

test/data/operator.ts
alfetopito: fake data, dont care =P

src/utils/format.ts
alfetopito: I'm capitalizing only the `kind`, not the whole sentence. As per design.

W3stside (commented):
approve

src/utils/format.ts
anxolin: Yes, I agree, I was seeing it correct in the UI. My comment was more like, that by only looking at the code, I was thinking u were capitalizing all (which I think it would be wrong), but obviously I read bad the code cause it's all good, so u can forget this comment ever existed :)

PR #184, opened 2021-02-19
Refactor block explorer link

alfetopito:
# Summary

Current `BlockExplorerLink` implementation relies on `useWalletConnection` hook that in turn uses `walletApi`.

This PR:
- Turns `BlockExplorerLink` into `DumbBlockExplorerLink`, removing the hooks
- Creates one `BlockExplorerLink` for `gp-v1`, with the same behaviour as before
- Creates one `BlockExplorerLink` for `explorer`, using the app's `network` state


src/apps/explorer/components/common/BlockExplorerLink/index.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+
+import { DumbBlockExplorerLink, Props } from 'components/common/DumbBlockExplorerLink'
```
anxolin, line 3: I don't think this is App specific

Any app using the same `state/network` could use this smart component.

anxolin (commented):
I like the core idea of this PR, making the dumb version of the Link. 
I'm not sure about names and structure. 

Some ideas:
- I wouldn't call the dumb component "Dumb", we haven't use this naming convention before, and we have a lot of dumb components already without pre/sufixes 
- For the dumb, I would call it BlockExplorerLink
- We make one BlockExplorerLink per app, that's not bad idea. Just for simplifying the names, and so we don't need to think in each app wich flavour of the smart one need
- BUT the implementation of it should be a common component, since they can be highly reused by other apps
- So Then we can have in `src/components/common/BlockExplorerLink`:
     - BlockExplorerLink --> dumb
     - BlockExplorerLinkNetworkStateWidget --> I hate the name, but it wouldn't matter since each app will rename it to `BlockExplorerLink`
     - BlockExplorerLinkWalletHookWidget --> same as above

anxolin (approved):
❤️

PR #196, opened 2021-02-22
Copy button component

alfetopito:
# Summary

Implements copy button component as per design
screenshot_2021-02-22_13-49-45

- Simple interface, takes only the text to be copied
- Icon color is theme dependent (light/dark) matching design
- On copy, icon changes to a green check for 2s, then back to original icon

# Testing

Storybook: https://pr196--gpui.review.gnosisdev.com/storybook/?path=/story/common-copybutton--default


src/components/common/CopyButton/CopyButton.stories.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export default {
+  title: 'Common/CopyButton',
+  component: CopyButton,
+  decorators: [GlobalStyles, ThemeToggler],
+} as Meta
+
+const Template: Story<Props> = (args) => (
+  <div style={{ fontSize: '15px' }}>
+    Click the button to copy
```
anxolin, line 17: uber-mega-super-nit: can u add a &nbsp; or a {' '}

src/components/common/CopyButton/index.tsx
```diff
@@ -0,0 +1,48 @@
+import React, { useEffect, useState } from 'react'
+import styled from 'styled-components'
+import { faCopy } from '@fortawesome/free-regular-svg-icons'
+import { faCheck } from '@fortawesome/free-solid-svg-icons'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import CopyToClipboard from 'react-copy-to-clipboard'
+
+import { DISPLAY_TEXT_COPIED_CHECK } from 'apps/explorer/const'
+
+const Icon = styled(FontAwesomeIcon)<{ copied: boolean }>`
```
anxolin, line 10: I'm fined with this name, but I want to make sure we are consistent. Do we have a naming convention for the styled components props. Should it be `$copied` or `copied`?

 @W3stside?

src/components/common/CopyButton/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+  color: ${({ theme, copied }): string => (copied ? theme.green1 : theme.icon)};
```
anxolin, line 11: @biocom u had some concerns with the theming. Is this subject settled?

anxolin (approved):

src/components/common/CopyButton/index.tsx
biocom: @alfetopito Appreciate the fact you implemented this, by going the extra mile :) - Only thing to add is to show 'copied' next the to green check on click (Like Etherscan) for explicit textual validation.

src/components/common/CopyButton/index.tsx
biocom: Not a concern, but would say that we'd likely benefit from having a single `green` color variable throughout. I know that right now we still support other UI's/codebase, but just stating as a goal.

anxolin: ok, although in this case I'm referring to some concerns about the theming system and the template literals. I mean, do we have consensus in how we style our theme?

biocom: To be honest I haven't dug to deep yet into how @W3stside setup theming. AFAIK we were using this method above Leandro implemented. But leaving this to David to correct where I'm off.

anxolin: okok, if u have no issues, then the matter is settled. 2 weeks ago, u expressed some concerns about the theme, so I was referring to that.

biocom: My concern there was specifically how we were writing the media queries :)

anxolin: ahh, right! only for media queries. OK. Are u still concern about that? should be settle it?

biocom: Yes, we should. Forgot to mention in our last sync. I will write on our channel.

src/components/common/CopyButton/index.tsx
alfetopito: > Should it be $copied or copied?

Well, not a big fan of `$` (jQuery feelings).
Anyway, it's an internal prop, so it's explicit enough.
Maybe if it was an exposed component it would be better do differentiate the props, but I don't see the need here.

> Only thing to add is to show 'copied' next the to green check on click (Like Etherscan) for explicit textual validation.
Sure, will add it

biocom: $(thankYou) $(veryMuch)

src/theme/styles/colours.ts
```diff
@@ -104,6 +108,8 @@ export const DARK_COLOURS = {
   textActive1: '',
   textDisabled: '#31323E',
 
+  icon: '#8D8DA980',
```
biocom, line 111: Why not use rgba() ?

src/components/common/CopyButton/index.tsx
W3stside: i think we decided a little while ago (too long to remember so not pointing fingers) that we are dropping the styled `$` convention as it's not super helpful AND clashes with `styled-theming`

src/theme/styles/colours.ts
alfetopito: Because all other colors are defined as `hex`. Wanted to keep the same pattern.
I don't mind changing, though. Do you prefer that?

src/components/common/CopyButton/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Simple CopyButton component.
+ *
+ * Takes in the `text` to be copied when button is clicked
+ * Displays the copy icon with theme based color
+ * When clicked, displays a green check for DISPLAY_TEXT_COPIED_CHECK seconds,
+ * then is back to original copy icon
+ */
+export function CopyButton(props: Props): JSX.Element {
+  const { text } = props
```
W3stside, line 26: not a fan of destructuring in signature are you :joy: :D

src/theme/styles/colours.ts
W3stside: haha the perfect "designer vs developer" situation - hex is much easier for us design-lacking devs to copy and paste, but i totally agree rgba is better for manipulating colours, and adds a nice opacity alpha. In this case it's def us keeping a standard format. 

We may want to consider switching entirely to rgba @alfetopito @anxolin

W3stside (approved):
approved af, nice work

src/components/common/CopyButton/index.tsx
alfetopito: Nope, I like the signature cleaner :)

src/theme/styles/colours.ts
anxolin: I think here @biocom should be the last authority. Whatever he feels more comfortable with. 
I personally don't modify ever a color by hand, I use a color picker. So hex is fine for me.

src/components/common/CopyButton/index.tsx
anxolin: I'm with @alfetopito in this one, with one param I find it okeish, but I also don't like this tough-guy-lines with declarations, arrow functions, ternary operators and || && ones. Yes, when you write them u feel u are a hacker and is clear, but the next one reading that, won't have such an easy time

1 thing per line as a general rule, sounds good to me. Of Course with exceptions

src/theme/styles/colours.ts
biocom: > haha the perfect "designer vs developer" situation - hex is much easier for us design-lacking devs to copy and paste, 

Besides copy and paste, does it provide you any context on opacity? Looking at:
`rgb(141 141 169 / 20%)`
`rgb(141 141 169 / 50%)`

I know I'm dealing with a single color, but with a different alpha (opacity) setting. Looking at a matching HEX example:
`#8d8da933`
`#8d8da980`

Can you honestly tell what the exact opacity is?

alfetopito: I personally cannot, my hex math is not the best =P

I would then be in favor of making all colors RGB[A], for no other reason than I can search for a color and know whether that's defined already somewhere or not. 
For example when I need a color and I don't know whether we have the same exact one, I do a text search for the hex value.

If that's alright, I can make a PR transforming the colors from hex to rgb[a]

biocom: My take would be:

1. Where ALPHA colors are not needed, just use a 6 digit HEX.
2. Where ALPHA is used, use the alpha version.

Ideally these colors are only defined in the `src/theme/styles/colours.ts` file, so searching for a color most of times would be by variable name?

That being said, while working on a front-end ticket myself, I was thinking to try to exclusively only use 6 digit HEXes and instead try to use the `opacity` CSS property where possible. 

As you know in some cases that won't work (e.g. needing transparency only on a `border` property without effecting the whole element with `opacity`).

PR #197, opened 2021-02-23
167/order surplus row

### LINKED ISSUE #167: [order details] Implement Order Surplus row
Order Surplus:
- % over requested price
- amount in surplus token (buy token when sell order, sell token when buy order)
- approximated USD amount

alfetopito:
# Summary

Closes #167 

Order surplus
screenshot_2021-02-22_17-23-51

- Unit tests for surplus calculation
- Storybook for OrderSurplusDisplay
- :warning: USD amount not implemented (out of the sprint)

# Testing:
- Storybook https://pr197--gpui.review.gnosisdev.com/storybook/?path=/story/orders-ordersurplusdisplay--surplus-100-percent
- Check a matched order (example for rinkeby https://pr197--gpui.review.gnosisdev.com/rinkeby/orders/0xa6223509c0a41f992995d8781c3452ae8e2f55640b21d057fc02c7fb5a1674b35b0abe214ab7875562adee331deff0fe1912fe4260246928)


src/api/operator/utils.ts
```diff
@@ -68,6 +68,28 @@ export function getOrderFilledAmount(order: RawOrder): { amount: BigNumber; perc
   return { amount: executedAmount, percentage: executedAmount.div(totalAmount) }
 }
 
+export function getSurplus(
```
anxolin, line 71: this function is pure, why here in the "operator" and not in utils?

src/api/operator/utils.ts
```diff
@@ -68,6 +68,28 @@ export function getOrderFilledAmount(order: RawOrder): { amount: BigNumber; perc
...
+  inputAmount: BigNumber | string,
+  executedAmount: BigNumber,
```
anxolin, line 73: it's a bit strange you allow string for the input and not the executedAmount. Not biggy, I think u did it for not handling the case for executedAmount  and while for inputAmount was coming for free

src/components/orders/OrderSurplusDisplay/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const usdAmount = '55.555'
+
+  if (!surplusToken || surplusAmount.isZero()) {
+    return null
+  }
+
+  return (
+    <Wrapper>
+      <Surplus>
+        {formatSmart({ amount: surplusPercentage.multipliedBy('100').toString(10), precision: 0, decimals: 2 })}
```
W3stside, line 54: can we use a constant for decimals

src/components/orders/OrderSurplusDisplay/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+      </Surplus>
+      <span>
+        {formatSmart({
+          amount: surplusAmount.toString(10),
+          precision: surplusToken.decimals,
+          decimals: 8,
```
W3stside, line 60: same for this badboi

src/components/orders/OrderSurplusDisplay/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+          smallLimit: '0.00000001',
```
W3stside, line 61: and for our poor lonely guy here, constant. Unless we don't want this? :shrimp:

W3stside (approved):
just minor things about constants and magic numbers in smart formatting otherwise very nice!

src/components/orders/OrderSurplusDisplay/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+  }
+`
+
+const Surplus = styled.span`
+  ::before {
+    content: '+';
+  }
+  ::after {
+    content: '%';
+  }
```
anxolin, line 26: why is this better than doing `+{ formatSmart({... }) }%` in line 54

anxolin (approved):
Awesome PR. Specially I liked you splitted the pure function, from the "order" function, and that we have simple tests and storybook for this. ❤️

src/components/orders/OrderSurplusDisplay/index.tsx
alfetopito: Just because I know how to do that with CSS now :smile: 
But I also thought it looks cleaner
Would you rather have it like you suggest? I don't really mind.

src/api/operator/utils.ts
alfetopito: They are still strongly related. I'm using a single test file for both, for instance.

But ok, I give up =P
Will migrate this utils out of the api into regular utils

src/api/operator/utils.ts
alfetopito: Mostly because executed amount will always be a BigNumber (from the sources I have so far).
But you are right, will make both inputs the same type

src/components/orders/OrderSurplusDisplay/index.tsx
W3stside: its more declarative pt 2 but this is also good cuz CSS skillz +1

biocom: In general my approach with pseudo-elements is: Is it a layout element or actual content? In this case I think the % sign is part of the content and would make it part of the 'normal' text. Mainly for accessibility purposes too.

E.g. IMO would be different if we use it to add a '-' separator which I'd regard as a visual element.

alfetopito: Makes sense.
Will change in another PR.

alfetopito: https://github.com/gnosis/gp-ui/pull/203

anxolin (commented):
image

Issue #174, opened 2021-02-18
Tooltips for the Explorer


anxolin:
Design + Texts

biocom:
Design done. Text pending.

biocom:
Addressed in PR https://github.com/gnosis/gp-ui/pull/201

PR #201, opened 2021-02-24
Gp tooltips

biocom:
Screen Shot 2021-02-24 at 15 20 05

**Implemented:**
- Tooltip styles, respecting the dark/light mode (theme).
- Tooltip text content
- Added some missing fields in the table.

**Todo:** 
- Define light theme colors. Will address in a future PR.
- Address the 'Fill' property field, by adding a progress bar. Will be part of a separate PR.


src/components/InputWithTooltip.tsx
```diff
@@ -68,9 +68,7 @@ const InputWithTooltip: React.RefForwardingComponent<HTMLInputElement, InputProp
   return (
     <>
       <Input $error={showErrorStyle} {...props} {...finalTargetProps} />
-      <Tooltip bgColor={tooltipBgColor} {...tooltipProps}>
```
biocom, line 71: Instead of passing `bgColor` as a prop, now we base it on a `theme` variable (shared between dark/light mode).

src/components/Tooltip.tsx
```diff
@@ -194,12 +204,13 @@ interface HelpTooltipProps {
   tooltip: ReactNode
   placement?: Placement
   offset?: number
-  iconSize?: FontAwesomeIconProps['size']
```
biocom, line 197: Opted for simplifying the props, by assuming we'd use a single universal size for the tooltip icon. If we want to deviate, I'd propose we add specific icon sizes in our `theme` and handle that separately.

src/theme/styles/colours.ts
```diff
@@ -112,13 +116,15 @@ export const DARK_COLOURS = {
   textActive1: '#D96D49',
   textDisabled: '#31323E',
 
-  icon: '#8D8DA980',
+  icon: '#8D8DA9',
```
biocom, line 119: Opting here for using a single color for the icon (without Alpha). The using `opacity` to set on hover effects.

src/theme/styles/colours.ts
```diff
@@ -100,7 +104,7 @@ export const LIGHT_COLOURS = {
   labelBgOpen: '#77838F1A',
 
   // table & borders
-  borderPrimary: 'rgba(151, 151, 184, 0.3)',
+  borderPrimary: '#3A3A48',
```
biocom, line 107: Removing alpha, as I realized the transparency might possibly have future color/blending interference with other elements.

src/components/InputWithTooltip.tsx
biocom: Opting for simplifying this tooltip component, until we decide we need additional props?

src/components/common/TooltipLegacy.tsx
```diff
@@ -0,0 +1,240 @@
+import React, { ReactNode, CSSProperties, useCallback } from 'react'
```
biocom, line 1: Moved the legacy tooltip to the `components/common` folder as per @alfetopito 's suggestion.

anxolin: Why this file looks like a new file? surprised git didn't detect it as just a rename. And shouldn't be somewhere the deletion of the old file? I can't find it

anxolin (approved):
Nice PR, and I like the messages too

src/components/orders/DetailsTable/index.tsx
```diff
@@ -49,6 +50,28 @@ const Table = styled(SimpleTable)`
...
+    'An order can be either a Buy or Sell order. In addition, an order may be of type "Fill or Kill" (no partial fills) or a regular order (partial fills allowed).',
+  sell: 'The total sell amount, denominated in the sell token.',
+  buy: 'The total buy amount, denominated in the buy token.',
+  priceLimit:
+    'The limit price is the price at which this order shall be (partially) filled, in combination with the specified slippage.',
+  priceExecution: 'The actual price at which this order has been matched and executed.',
+  surplus:
+    'The (averaged) surplus for this order. This is the positive difference between the initial limit price and the actual (average) execution price.',
+  filled: 'Indicates what percentage amount this order has been filled and the amount sold/bought.',
+  fees: 'The amount of fees paid for this order. This will show a progressive number for orders with partial fills.',
```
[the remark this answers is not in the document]
biocom: I meant to convey the fact that, depending on the order, it would show a progressive number. But I can see your argument for keeping it simple.

src/components/orders/DetailsTable/index.tsx
```diff
@@ -49,6 +50,28 @@ const Table = styled(SimpleTable)`
...
+  submission:
+    'The date and time at which the order was submitted. The timezone is based on the browser locale settings.',
+  expiration:
+    'The date and time at which an order will expire and effectively be cancelled. Depending on the type of order, it may have partial fills upon expiration.',
+  type:
+    'An order can be either a Buy or Sell order. In addition, an order may be of type "Fill or Kill" (no partial fills) or a regular order (partial fills allowed).',
+  sell: 'The total sell amount, denominated in the sell token.',
+  buy: 'The total buy amount, denominated in the buy token.',
+  priceLimit:
+    'The limit price is the price at which this order shall be (partially) filled, in combination with the specified slippage.',
```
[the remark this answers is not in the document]
biocom: I guess maybe too much information? But wanted to convey that this is the limit price, but keeping in mind the slippage setting too.

src/components/common/TooltipLegacy.tsx
biocom: Because the old file name was Tooltip.tsx, and I use that name as our current Tooltip component. So I copied the previous Tooltip.tsx and renamed it to --> TooltipLegacy.tsx

biocom:
Updated to resolve a merge conflict with PR 205 @alfetopito @anxolin

anxolin:
closes #174

src/components/common/TooltipLegacy.tsx
alfetopito: Actually, my suggestion is to move this further away, keeping the same name.
Instead of `src/components/common/TooltipLegacy`, I'd rather move it to `src/apps/gp-v1/components/common/Tooltip`.
And update `gp-v1` references to point to the same tooltip component but with a different path now.

src/components/InputWithTooltip.tsx
alfetopito: This component is also legacy. I would rather move it to `apps/gp-v1/components/common` as well and use the legacy tooltip component to avoid any issues we might have if/when we decide to release a new version of the old interface.

biocom: Ah I see. Good if I address this in this PR?

src/components/common/TooltipLegacy.tsx
biocom: Ah I see your point, makes perfect sense. Will address in this PR.

src/components/InputWithTooltip.tsx
alfetopito: Sure, I don't see a problem doing it here.

PR #205, opened 2021-02-24
168/amount row

### LINKED ISSUE #168: [order details] Implement Amount row
Amount

- two rows: buy on top, sell on bottom
- estimate in USD
- images for each token
- link to each token on etherscan
- Handle From (to at least), or To (to at least) depending if it's buy order and sell order

alfetopito:
# Summary

Closes #168 

- Does not yet have USD estimation. To be addressed in another task
- Images are most likely to work on mainnet. Additional work to be done in another task.
  For now re-using the same component used for Mesa
- Storybook does not have the network state, thus no link is displayed (:disappointed:)

## Rinkeby - on the page
screenshot_2021-02-24_15-14-19
## Storybook - mocked data
screenshot_2021-02-24_15-14-34

### DIFF
--- a/src/components/orders/AmountsDisplay/index.tsx
+++ b/src/components/orders/AmountsDisplay/index.tsx
@@ -1,0 +1,72 @@
+import React from 'react'
+import BigNumber from 'bignumber.js'
+
+import { formatSmart, TokenErc20 } from '@gnosis.pm/dex-js'
+
+import { Order } from 'api/operator'
+
+import { HIGH_PRECISION_DECIMALS, HIGH_PRECISION_SMALL_LIMIT } from 'apps/explorer/const'
+
+import { BlockExplorerLink } from 'apps/explorer/components/common/BlockExplorerLink'
+
+import { RowContents, RowTitle, StyledImg, UsdAmount, Wrapper } from './styled'
+
+type RowProps = {
+  title: string
+  showTitleSuffix: boolean
+  amount: BigNumber
+  erc20: TokenErc20
+}
+
+function Row(props: RowProps): JSX.Element {
+  const { title, showTitleSuffix, amount, erc20 } = props
+
+  // TODO: calculate real usd amount
+  const usdAmount = '31231.32'
+
+  // Decimals are optional on ERC20 spec. In that unlikely case, graceful fallback to raw amount
+  const formattedAmount = erc20.decimals
+    ? formatSmart({
+        amount: amount.toString(10),
+        precision: erc20.decimals,
+        decimals: HIGH_PRECISION_DECIMALS,
+        smallLimit: HIGH_PRECISION_SMALL_LIMIT,
+      })
+    : amount.toString(10)
+  // Name and symbol are optional on ERC20 spec. Fallback to address when no name,
+  // and show no symbol when that's not set
+  const tokenLabel = `${erc20.name || erc20.address}${erc20.symbol ? ` (${erc20.symbol})` : ''}`
+
+  return (
+    <>
+      <RowTitle>
+        {title} {showTitleSuffix && 'at least'}
+      </RowTitle>
+      <RowContents>
+        <span>{formattedAmount}</span>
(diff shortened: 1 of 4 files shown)


src/components/orders/AmountsDisplay/styled.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export const RowContents = styled.span`
+  display: flex;
+  align-items: center;
+
+  & > :first-child {
+    margin: 0 0.5rem 0 0;
+  }
+
+  & > * {
```
biocom, line 32: I'd prevent the wildcard at all times, because of performance and unexpected matches. 

I'd be as specific in the structure as possible (like you did with & > :first-child).

src/components/orders/AmountsDisplay/styled.tsx
```diff
@@ -0,0 +1,44 @@
...
+  }
+`
+
+export const UsdAmount = styled.span`
+  color: ${({ theme }): string => theme.bgDisabled};
+`
+
+export const StyledImg = styled(TokenImg)`
+  width: 16px;
+  height: 16px;
```
biocom, line 43: 16px -> 1.6rem ?

biocom:
Small comment: The tokens are orange links. In the design I opted to only have the underline effect 'on hover'.

anxolin:
Added https://github.com/gnosis/gp-ui/issues/207 to tackle what u mentioned in the description

src/components/orders/AmountsDisplay/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+import { HIGH_PRECISION_DECIMALS, HIGH_PRECISION_SMALL_LIMIT } from 'apps/explorer/const'
+
+import { BlockExplorerLink } from 'apps/explorer/components/common/BlockExplorerLink'
+
+import { RowContents, RowTitle, StyledImg, UsdAmount, Wrapper } from './styled'
+
+type RowProps = {
+  title: string
+  showTitleSuffix: boolean
```
anxolin, line 16: nit: This is for the `at least`, I would just call it `amountIsLowerBound` or sth in those lines. Or do sth like `amountIsFixed`, or `amountIsAtMost`, or `amountIsAtLeast`

if you want to call that title suffix, I would expect it to be an optional suffix string. But I think,

anxolin (approved):
Approving. I haven't reviewed the CSS too much which I think michel can give a better review

biocom (approved):

src/components/orders/AmountsDisplay/index.tsx
alfetopito: You cut off in the middle, but I get the point.
Forgot that it's `at most` when selling and  `at least` when buying.
Think I'll instead take the suffix as a parameter instead of a boolean.
New PR coming soon

src/components/orders/AmountsDisplay/styled.tsx
alfetopito: In this case, what would be the best way to select all children without using wild cards?
I assume an inverted selector would be equally bad, such as: `& > :not(:first-child)` ?

src/components/orders/AmountsDisplay/index.tsx
alfetopito: https://github.com/gnosis/gp-ui/pull/209

alfetopito:
> Small comment: The tokens are orange links. In the design I opted to only have the underline effect 'on hover'.

Ahhh I see what you mean now. Got confused there for a minute.
Will do that in the follow up PR https://github.com/gnosis/gp-ui/pull/209

src/components/orders/AmountsDisplay/styled.tsx
biocom: My first thought would be to try to use a single HTML element you could target. If not possible specify all selectors?

PR #206, opened 2021-02-25
Use inter-ui font import + variable font & cleanup.

biocom:
Related issue #68 

This PR cleans up the existing fonts and replaces it with an import of `inter-ui`. This tries to favor the use of a variable font type. Similar approach to Uniswap's front-end.

- Cleanup old fonts/folders
- Use `inter-ui` font import
- Check support for variable fonts and use a fallback for a normal font type.
- Cleanup of font variables. Progressively add more variables for defining font sizes etc.


anxolin (approved):
👌

alfetopito (changes requested):
You removed some font types, but they were in use, which caused the build to break

screenshot_2021-02-25_10-11-14

biocom:
> You removed some font types, but they were in use, which caused the build to break

I know I was a little bold by removing it all in this PR :)

My assumption was this will not be used in any of our new apps. But I understand we still want legacy support at least for the older interface? I will address this in this PR (reverting back those files).

alfetopito:
> > You removed some font types, but they were in use, which caused the build to break
> 
> I know I was a little bold by removing it all in this PR :)
> 
> My assumption was this will not be used in any of our new apps. But I understand we still want legacy support at least for the older interface? I will address this in this PR (reverting back those files).

@biocom not a problem at all to remove them.

Just pointing out to know what needs to be updated, since you are pretty much renaming theme properties.
I think updating the three files mentioned in the error should fix the issue.

You can confirm locally by running `yarn build`

And this theme is used only in the new apps AFAIK. The errors happened only in the new components.

biocom:
@alfetopito Addressed the issue with renaming font variables, which caused the build to fail.

alfetopito (approved):
Awesome :+1:

annamsgeorge:
Detail: 
Lets reword to:

"Protocol/Gas fee"

And in the hover text:
"All fees (incl. gas fees) are already included in your receiving balance. You don't need ETH for this trade."

PR #213, opened 2021-02-26
Fix non-boolean attribute warning for CopyButton

alfetopito:
# Summary

Fixes non-boolean HTML error/warning:
screenshot_2021-02-26_09-54-13

# Testing
1. Open the order details page
2. Open the console
3. Verify the error mentioned above is not there

### DIFF
--- a/src/components/common/CopyButton/index.tsx
+++ b/src/components/common/CopyButton/index.tsx
@@ -7,7 +7,13 @@ import CopyToClipboard from 'react-copy-to-clipboard'
 
 import { DISPLAY_TEXT_COPIED_CHECK } from 'apps/explorer/const'
 
-const Icon = styled(FontAwesomeIcon)<{ copied: boolean }>`
+// Why is `copied` not a boolean?
+//   Because it's passed down to parent component (`FontAwesomeIcon`) and
+// since it's not consumed in the component, it's passed down to the HTML.
+//   Which does not like to receive boolean values, with an error like:
+// "Warning: Received `false` for a non-boolean attribute `copied`"
+//   Effectively though, it's treated as a boolean, thus the value doesn't matter
+const Icon = styled(FontAwesomeIcon)<{ copied?: string }>`
   color: ${({ theme, copied }): string => (copied ? theme.green : theme.grey)};
   transition: color 0.2s ease-in;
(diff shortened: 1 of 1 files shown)


biocom (approved):

src/components/common/CopyButton/index.tsx
```diff
@@ -7,7 +7,13 @@ import CopyToClipboard from 'react-copy-to-clipboard'
 
 import { DISPLAY_TEXT_COPIED_CHECK } from 'apps/explorer/const'
 
-const Icon = styled(FontAwesomeIcon)<{ copied: boolean }>`
+// Why is `copied` not a boolean?
+//   Because it's passed down to parent component (`FontAwesomeIcon`) and
+// since it's not consumed in the component, it's passed down to the HTML.
+//   Which does not like to receive boolean values, with an error like:
+// "Warning: Received `false` for a non-boolean attribute `copied`"
+//   Effectively though, it's treated as a boolean, thus the value doesn't matter
```
anxolin, line 15: Thanks for explaining this :) 

But I mean, it feels weird to pass it as a string. On alternative I see, is just to add a className to the icon, when is copied. Then in the css you can just do:

```
. copied {
    cursor: ${({ copied }): string => (copied ? 'reset' : 'pointer')};
}
```

although is not super sytled-component-like, It looks pretty readable to me

anxolin (approved):
APPROVED, although, consider my comment if possible :)

src/components/common/CopyButton/index.tsx
alfetopito: Yeah, that could work.
Another alternative is to "consume" the boolean and prevent it from being passed down to the HTML.
I did that for some components on CMM app: https://github.com/gnosis/safe-cmm-app-react/blob/131f3197a4e746fae880241bb950fcbe576bd968/src/components/basic/display/BracketsTable/Table.tsx#L43

Which of the 3 options do you prefer?

anxolin: your's might be more pure style-components like, but still feels overkillin. I don't know, I find the `className` one pretty easy to read, and has the benefit that we don't have to use this interpolation functions

Not strong opinion, go with whatever you feel is right of the 3.

PR #231, opened 2021-03-01
Making the order detail view responsive.

biocom:
General PR to make the order detail table responsive.

### DIFF
--- a/src/components/common/SimpleTable/index.tsx
+++ b/src/components/common/SimpleTable/index.tsx
@@ -1,6 +1,6 @@
 import React from 'react'
 import styled from 'styled-components'
-import { applyMediaStyles } from 'theme'
+import { media } from 'theme/styles/media'
 
 const Wrapper = styled.table<{ $numColumns?: number }>`
   font-size: ${({ theme }): string => theme.fontSizeDefault};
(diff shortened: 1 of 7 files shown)


src/theme/styles/global.ts
```diff
@@ -19,29 +19,13 @@ const selection = css`
   /* End CSS for selecting text */
 `
 
-const scrollbars = css`
```
biocom, line 22: Scrollbars were being applied site wide. Revert back to browser defaults. Apply only in cases where we want to deviate from this (e.g. inner scrolling tables).

src/components/common/SimpleTable/index.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react'
 import styled from 'styled-components'
-import { applyMediaStyles } from 'theme'
+import { media } from 'theme/styles/media'
```
biocom, line 3: Using the previous media queries `media.ts` file.

anxolin (approved):
Awesome.

Leaving here a DEMO link to make it easier for the reviewers https://pr231--gpui.review.gnosisdev.com/rinkeby/orders/0x3960fcd923574dcd0c0f40f9bd20485a2c8b4f85128eedb3c3d53960691d1d7d5b0abe214ab7875562adee331deff0fe1912fe4260257f55/

alfetopito (approved):
That is quite nice :+1: 

One thing not related to your changes, but should be addressed is the footer:
screenshot_2021-03-01_11-03-20

Maybe add padding top/bottom? Although that will make the footing more prominent.

biocom:
> That is quite nice 👍
> 
> One thing not related to your changes, but should be addressed is the footer:
> screenshot_2021-03-01_11-03-20
> 
> Maybe add padding top/bottom? Although that will make the footing more prominent.

Planned to address the footer in a separate commit (plan to *not* make it position: fixed anymore too). This PR was solely for the order table. Thanks for mentioning.

PR #235, opened 2021-03-02
Adding fill progress bar.

biocom:
Adds a fill progress bar for the Filled section. Created a common component 'progressBar' to be re-used:

Screen Shot 2021-03-02 at 13 39 37
Screen Shot 2021-03-02 at 13 39 30

https://pr235--gpui.review.gnosisdev.com/storybook/?path=/story/common-progressbar--default
https://pr235--gpui.review.gnosisdev.com/storybook/?path=/story/orders-filledprogress--default

### DIFF
--- a/src/components/common/ProgressBar/index.tsx
+++ b/src/components/common/ProgressBar/index.tsx
@@ -4,6 +4,7 @@ import styled from 'styled-components'
 export type Props = {
   percentage?: string | null
   activeColor?: string
+  showLabel?: boolean
 }
 
 const Wrapper = styled.div<Props>`
@@ -28,14 +29,22 @@ const Wrapper = styled.div<Props>`
     background-color: ${({ activeColor, theme }): string => (activeColor ? activeColor : theme.green)};
     border-radius: 16rem;
   }
+
+  > b {
+    color: ${({ activeColor, theme }): string => (activeColor ? activeColor : theme.green)};
+    margin: 0 0 0 0.7rem;
+  }
 `
 
 export function ProgressBar(props: Props): JSX.Element {
+  const { percentage, showLabel = true } = props
+
   return (
     <Wrapper {...props}>
       <div>
         <span></span>
       </div>
+      {showLabel && <b>{percentage}%</b>}
     </Wrapper>
   )
 }
--- a/src/components/common/ProgressBar/index.tsx
+++ b/src/components/common/ProgressBar/index.tsx
@@ -2,9 +2,9 @@ import React from 'react'
 import styled from 'styled-components'
 
 export type Props = {
-  percentage?: string | null
-  activeColor?: string
-  showLabel?: boolean
+  readonly percentage?: string
+  readonly activeColor?: string
+  readonly showLabel?: boolean
 }
 
 const Wrapper = styled.div<Props>`
@@ -37,7 +37,7 @@ const Wrapper = styled.div<Props>`
 `
 
 export function ProgressBar(props: Props): JSX.Element {
-  const { percentage, showLabel = true } = props
+  const { percentage = '0', showLabel = true } = props
 
   return (
     <Wrapper {...props}>
--- a/src/components/common/ProgressBar/index.tsx
+++ b/src/components/common/ProgressBar/index.tsx
@@ -1,0 +1,41 @@
+import React from 'react'
+import styled from 'styled-components'
+
+export type Props = {
+  percentage?: string | null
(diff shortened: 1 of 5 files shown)


biocom:
Adressing #68

alfetopito (approved):
Great work, looks amazing!

src/components/common/ProgressBar/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+    margin: 0 0 0 0.7rem;
+  }
+`
+
+export function ProgressBar(props: Props): JSX.Element {
+  const { percentage = '0', showLabel = true } = props
+
+  return (
+    <Wrapper {...props}>
+      <div>
```
biocom, line 44: I kinda like having a compact styled component with very specific nested selectors (> b { ....)

But in terms of readability, would it be preferred to turn the nested elements into it's own styled components? E.g.

<div> --> OuterBar
<span> --> InnerBar

?

src/components/common/ProgressBar/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  > b {
+    color: ${({ activeColor, theme }): string => (activeColor ? activeColor : theme.green)};
+    margin: 0 0 0 0.7rem;
+  }
+`
+
+export function ProgressBar(props: Props): JSX.Element {
+  const { percentage = '0', showLabel = true } = props
```
biocom, line 40: Providing defaults, given we have optional props. Preferred way to do it this way?

src/components/common/ProgressBar/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+  > div {
+    width: 16rem;
+    height: 0.6rem;
+    position: relative;
+    border-radius: 16rem;
+    background: ${({ theme }): string => theme.borderPrimary};
+  }
+
+  > div > span {
+    width: ${({ percentage }): string => percentage || '0'}%;
```
biocom, line 23: How to keep the fallback out of this here? Can only be done in the parent component were we get the props from?

src/components/common/ProgressBar/index.tsx
anxolin: I'm happy with both approaches. By the size of this component, your approach is good. 
Probably u are right that using  it's own styled components is more readable, but it's kind of fine in this specific case.

src/components/orders/FilledProgress/FilledProgress.stories.tsx
```diff
@@ -0,0 +1,23 @@
...
+} as Meta
+
+const Template: Story<Props> = (args) => <FilledProgress {...args} />
+
+const defaultArgs: Props = {
+  percentage: '40',
+}
+
+export const Default = Template.bind({})
+Default.args = { ...defaultArgs }
```
anxolin, line 23: nit picking, I would suggest to add another two stories: 0% and 100%

anxolin (approved):
Looks great!

src/components/common/ProgressBar/index.tsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react'
+import styled from 'styled-components'
+
+export type Props = {
+  readonly percentage?: string
+  readonly activeColor?: string
+  readonly showLabel?: boolean
+}
+
+const Wrapper = styled.div<Props>`
```
anxolin, line 10: You only do this if it's true that you use all the props in the styled-component, but for example `showLabel` is not used

It's more declarative if you use the ones you need. And then, for example, you would be able to define `percentage` as mandatory, so you can do the `percentage || 0` in the reder function, not in the CSS

src/components/common/ProgressBar/index.tsx
anxolin: You can define the props of the styled compoenent different to the component itself. See my other comment

src/components/common/ProgressBar/index.tsx
anxolin: yes, perfect

src/components/common/ProgressBar/index.tsx
alfetopito: I personally prefer smaller styled components, but am ok with this approach too.

src/components/common/ProgressBar/index.tsx
biocom: Makes sense, thank you.

PR #236, opened 2021-03-02
Style header/networkLabel/footer.

biocom:
Styling several elements:
- Header + network label
- Footer

Screen Shot 2021-03-02 at 16 14 37
Screen Shot 2021-03-02 at 16 14 06

### DIFF
--- a/src/apps/explorer/layout/Header.tsx
+++ b/src/apps/explorer/layout/Header.tsx
@@ -7,29 +7,33 @@ import { useNetworkId } from 'state/network'
 import styled from 'styled-components'
 
 const Logo = styled.span`
-  display: block;
-  margin: 0 0.5rem;
-  font-size: 2rem;
+  display: flex;
+  align-items: center;
+  margin: 0 0.5rem 0 0;
+  font-size: 1.8rem;
+  line-height: 1;
+  font-weight: ${({ theme }): string => theme.fontBlack};
 `
 
 const NetworkLabel = styled.span`
-  display: block;
+  border-radius: 0.6rem;
+  display: flex;
   margin: 0 0.5rem;
   font-size: 1.1rem;
   text-align: center;
-  padding: 0.3rem 0.8rem;
+  padding: 0.7rem;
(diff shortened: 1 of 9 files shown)


alfetopito (commented):
Header looks great, as well as mobile footer.

One concern regarding desktop footer. It seems to hover neither attached to the bottom of the page nor the bottom of the content:

Home:
screenshot_2021-03-02_07-42-58

Order details:
screenshot_2021-03-02_07-42-42

I would expect it to bind to the bottom of the window as before and not have the empty space, more like etherscan:
screenshot_2021-03-02_07-51-38

Hmm but looking at Mesa this change has the same style:
screenshot_2021-03-02_07-52-01

I personally prefer footer at the foot (:wink:), but am no designer, so, no strong opinion.

biocom:
@anxolin @W3stside @alfetopito Also on another note: These changes will have impact on the classic trading UI. Right know I don't know if/when this classic trading UI will happen. 

Therefore I'd prioritize our current common components to be compatible with the Explorer + Swap UI. This means that some components (like the header) cause visual artifacts on the classic trading UI.

alfetopito (approved):
Alright, I'm ok with it then.

And I don't think we should worry too much about the trade app right now.
We can address any issues when we get to it.

W3stside:
> @alfetopito The approach here is to not bind it to anything. Rather just follow a directional top/down flow. Looking at the Etherscan example, the footer is just a block (no position fixed/absolute).
> 
> Pages with minimal content will need to be dealt with individually (e.g. by setting a fixed min-height). But ideally we'd not set this artificial height if we can.
> 
> I anticipate both the Explorer and Swap UI to have sufficient (vertical) content where this shouldn't be an issue.
> 
> As for the footer itself, long-term we can build the footer out (akin to Etherscan) so that it will occupy a bit more vertical space.

can't we just `margin-top: auto` to auto adjust it to the bottom with `height: 100%` set on the `body` anyways?

W3stside (approved):

PR #239, opened 2021-03-02
[Analytics] Add copy event

anxolin:
Adds tracking event support 

The analytics API expose a triggerEvent.
It's used to track when the user copies the address, orderId and tx from the orders page


alfetopito (approved):

src/components/orders/DetailsTable/index.tsx
```diff
@@ -107,6 +110,13 @@ export function DetailsTable(props: Props): JSX.Element | null {
     return null
   }
 
+  const onCopy = (label: string): void =>
+    triggerEvent({
+      category: 'Order details',
+      action: 'Copy',
```
biocom, line 116: @anxolin Might want to be more specific about what is being copied here. E.g. 'Copy tx hash'  or 'Copy account address'.

src/components/common/CopyButton/index.tsx
```diff
@@ -40,10 +40,13 @@ export type Props = { text: string }
  * then is back to original copy icon
  */
 export function CopyButton(props: Props): JSX.Element {
-  const { text } = props
+  const { text, onCopy: onCopyProp } = props
```
W3stside, line 43: nit: onCopyProp is weird, we usually name handlers as such; `handleOnCopy`

src/components/orders/DetailsTable/index.tsx
W3stside: isn't that what `category` is for?

src/components/orders/RowWithCopyButton/index.tsx
```diff
@@ -17,17 +17,18 @@ type Props = {
...
   className?: string
+  onCopy?: (value: string) => void
 }
 
 export function RowWithCopyButton(props: Props): JSX.Element {
-  const { textToCopy, contentsToDisplay, className } = props
+  const { textToCopy, contentsToDisplay, className, onCopy } = props
 
   // Wrap contents in a <span> if it's a raw string for proper CSS spacing
   const contentsComponent = typeof contentsToDisplay === 'string' ? <span>{contentsToDisplay}</span> : contentsToDisplay
```
W3stside, line 27: this is weird, why not always render the span?

W3stside (approved):
approved w/comments

src/components/orders/DetailsTable/index.tsx
anxolin: I'm being specific. I specify that in the `label` : orderId / tx / address

anxolin: I did a small change so the labels are now: `orderId` / `settlementTx` / `ownerAddress`

src/components/orders/RowWithCopyButton/index.tsx
anxolin: Not from this PR, according to the comment is for styling. I agree, it feels a bit strange. 
Maybe @alfetopito wants to add sth.

anxolin: I looks like it can be `string | JSX.Element`, and it's just wrapping the element in case of being a string

I think is just fine...

alfetopito: Because:

1. The contents can be a string. In that case, wrap it in a `<span>` for CSS alignment
2. The contents can also be a component. In that case, there's no need to wrap it

PR #240, opened 2021-03-03
Use a different infura id for explorer

anxolin:
This PR changes the infura key being used. 
Now there's one specific for the Explorer. We leave the gp-v1 just for the legacy app in it's own config.

### DIFF
--- a/config-default.yaml
+++ b/config-default.yaml
@@ -76,7 +76,7 @@ defaultProviderConfig:
   type: 'infura' # Choices: infura | url
   config:
     # It'll be appended to `infuraEndpoint`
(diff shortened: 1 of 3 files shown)


W3stside (approved):
Approved but I think it's a bit sus to use a public infuraId as some rando can just us it and rate limit our apps.. or am i wrong?

alfetopito (approved):

anxolin:
> Approved but I think it's a bit sus to use a public infuraId as some rando can just us it and rate limit our apps.. or am i wrong?

This is always the case, and there's a functionality in infura for only allowing some origins

PR #242, opened 2021-03-03
Analytics: Add external link support

anxolin:
WIP: 

Adds a new component `OutboundLink`  for links that take you outside of the app, so we track when the users use it.

Thanks  @W3stside, @alfetopito for the help

I didn a small modifications to be as close as possible as the signature of the tative `<a />` element.  So now `to` is called `href`, and we don't required to add the `eventLabel` (we use the link by default)

This PR adds the ExternalLink component, in the 3 links in the footer, plus in the BlockExplorer component.


src/components/analytics/ExternalLink.tsx
```diff
@@ -0,0 +1,23 @@
+export { OutboundLink as default } from 'react-ga'
+
+// import React, { PropsWithChildren, useCallback } from 'react'
+// import { OutboundLink, OutboundLinkProps } from 'react-ga'
+
+// type OutboundLinkPropsWithOptionalLabel = Exclude<OutboundLinkProps, 'eventLabel'> & {
```
alfetopito, line 6: Shouldn't be `Omit` instead of `Exclude`?

alfetopito: Maybe try this:

```suggestion
type OutboundLinkPropsWithOptionalLabel = Omit<OutboundLinkProps, 'eventLabel'> & Partial<Pick<OutboundLinkProps, 'eventLabel'>>
```
```

File: src/components/analytics/ExternalLink.tsx [hunk 2]
```diff
@@ -0,0 +1,24 @@
+import React, { PropsWithChildren, ReactNode } from 'react'
+import { OutboundLink, OutboundLinkProps } from 'react-ga'
+
+type PropsOriginal = PropsWithChildren<OutboundLinkProps & React.HTMLProps<HTMLAnchorElement>>
+type Props = Omit<PropsOriginal, 'ref' | 'to' | 'eventLabel'> & {
+  eventLabel?: string
```
anxolin, line 6: I deliberately chosen to redefine this field by hand instead of using helpers like `Partial<Pick<OutboundLinkProps, 'eventLabel'>>`, I think it reads easier so I prefered to not make a TS helper hell

Same thing for the children

W3stside:
@anxolin I'm confused, my change I did locally worked fine, what did you try?

src/components/analytics/ExternalLink.tsx
alfetopito: You can make intermediary types to:
1. Avoid re-creating types
2. Increase readability

Like `PropsOriginal`

But I won't block you on this :)

alfetopito (approved):

src/components/analytics/ExternalLink.tsx
anxolin: Yes, I can make a 2nd intermediate types. My point, is I feel is overkilling. I kind of find this setup a good balance

I just feel is easier to read `eventLabel?: string`, thant adding an intermediate `type with Partial<Pick<OutboundLinkProps, 'eventLabel'>>`

W3stside (approved):

PR #244, opened 2021-03-03
232/filled row logic

### LINKED ISSUE #232: [order details] Filled row
Implement `Filled` row as per design

## Partially filable

No differences between the types

alfetopito:
# Summary

Closes #232 

Building up on the UI added by Michel, this PR wires up the logic

# Testing

- Check any order against the PR build
- Check the storybook


src/components/orders/FilledProgress/index.tsx
```diff
@@ -26,13 +33,71 @@ const Wrapper = styled.div`
...
+      <ProgressBar percentage={formattedPercentage} />
       <span>
-        <b>2,430 DAI</b> of <b>3000 DAI</b> sold for a total of <b>2.842739643 ETH</b>
+        <b>
+          {formattedFilledAmount} {mainSymbol}
+        </b>{' '}
+        of{' '}
+        <b>
+          {formattedMainAmount} {mainSymbol}
+        </b>{' '}
```
alfetopito, line 96: Question for the reviewer: should we always show both amounts, or only when not 100% filled?
screenshot_2021-03-03_11-08-31

src/components/orders/FilledProgress/index.tsx
```diff
@@ -26,13 +33,71 @@ const Wrapper = styled.div`
...
+    action = 'bought'
+  }
+
+  // In case the token object is empty, display the address
+  const mainSymbol = mainToken ? safeTokenName(mainToken) : mainAddress
+  const swappedSymbol = swappedToken ? safeTokenName(swappedToken) : swappedAddress
+  // In case the token object is empty, display the raw amount (`decimals || 0` part)
+  const formattedFilledAmount = formatSmart(filledAmount.toString(10), mainToken?.decimals || 0)
+  const formattedMainAmount = formatSmart(mainAmount.toString(10), mainToken?.decimals || 0)
+  const formattedSwappedAmount = formatSmart(swappedAmount.toString(10), swappedToken?.decimals || 0)
```
alfetopito, line 82: Question for the reviewer: Should we have greater precision here?

Right now it'll show `< 0.001` for anything < `0.001`...

Since the amounts here are redundant with info elsewhere I chose not to change this initially.
Have nothing against making it more precise, though.

src/components/orders/FilledProgress/index.tsx
biocom: I'd say showing both amounts feels more consistent. Although I could see only showing the full amount when it's fully filled.

src/components/orders/FilledProgress/index.tsx
biocom: Might indeed be relevant for certain tokens. Although you could opt to address this at a later point. Unless the level of efforts is low to support greater precision.

alfetopito: This is quite easy do add. Just a matter of what should we default to

anxolin (approved):

src/components/orders/FilledProgress/index.tsx
W3stside: i mean 0 seems low. considering it's sth like 18 (from token.decimals) else 0. Maybe do default 4 or something

W3stside (approved):

src/components/orders/FilledProgress/index.tsx
alfetopito: Default to `0` is only when the token is not set. So it "gracefully" handles when there's no token information and shows the raw amount.
This shouldn't happen anyway, as the parent component will not render when that's missing.
Added just for the sake of completeness. The alternative is to return `null` and not render the component at all.

Anyway, that's not what I was asking for. It's whether the format `decimals` (different than `precision`) should be higher.

anxolin: In general, for the explorer I don't like too much the `< 0.001` thing. Reason, is because it exposes the info from the protocol, as informative as possible. Even `0.001` can be worth something, if we talk about BTC it's almost 50$

If it's clearly redundant, maybe is fine. But only if it's very clear to the user.

alfetopito: ook, will update in another PR

alfetopito: https://github.com/gnosis/gp-ui/pull/280

PR #253, opened 2021-03-05
Upgrade ts

anxolin:
supersedes https://github.com/gnosis/gp-ui/pull/221 

Fixes an issue. We were using the spread array syntax that is available in newer versions of tslib.

image

Added the dev dependency with a modern version of tslib


alfetopito (approved):

W3stside (approved):
bom trabalho

PR #257, opened 2021-03-05
Was using wrong amounts to show the filled amount

alfetopito:
I noticed there was a surplus, but didn't seem like what was bought/sold had any surplus

Pay attention to the `Filled` row total on both cases:

## Before
screenshot_2021-03-05_13-57-57
## After
screenshot_2021-03-05_13-58-49

The second one shows the buy amount + surplus


anxolin (approved):
Great catch!

W3stside (approved):

PR #256, opened 2021-03-05
251/update contract address

alfetopito:
# Summary

Closes #251

Updating explorer contract address

Added both Settlement and Allowance manage addresses to footer
@biocom please review placement/wording

## Mobile
screenshot_2021-03-05_13-24-31
## Condensed
screenshot_2021-03-05_13-24-22
## Fully expanded
screenshot_2021-03-05_13-24-12

PR #279, opened 2021-03-08
Pr256/comments

alfetopito:
Addressing comments from https://github.com/gnosis/gp-ui/pull/256

### DIFF
--- a/src/api/deposit/batchExchangeAddresses.ts
+++ b/src/api/deposit/batchExchangeAddresses.ts
@@ -3,9 +3,9 @@ import { Network } from 'types'
 import { getGpV1ContractAddress } from 'utils/contract'
 
 const addressesByNetwork = {
-  [Network.Rinkeby]: getGpV1ContractAddress(Network.Rinkeby) || '',
-  [Network.Mainnet]: getGpV1ContractAddress(Network.Mainnet) || '',
-  [Network.xDAI]: getGpV1ContractAddress(Network.xDAI) || '',
+  [Network.Rinkeby]: getGpV1ContractAddress(Network.Rinkeby),
+  [Network.Mainnet]: getGpV1ContractAddress(Network.Mainnet),
+  [Network.xDAI]: getGpV1ContractAddress(Network.xDAI),
 }
 
 export const getAddressForNetwork = (networkId: Network): string | null => {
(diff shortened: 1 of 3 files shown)


src/api/deposit/batchExchangeAddresses.ts
```diff
@@ -3,9 +3,9 @@ import { Network } from 'types'
 import { getGpV1ContractAddress } from 'utils/contract'
 
 const addressesByNetwork = {
-  [Network.Rinkeby]: getGpV1ContractAddress(Network.Rinkeby) as string,
-  [Network.Mainnet]: getGpV1ContractAddress(Network.Mainnet) as string,
-  [Network.xDAI]: getGpV1ContractAddress(Network.xDAI) as string,
+  [Network.Rinkeby]: getGpV1ContractAddress(Network.Rinkeby) || '',
```
W3stside, line 6: should the function just return empty string? anyways i think you know what you're doing here more than me so it's probably fine

src/components/layout/GenericLayout/Footer/index.tsx
```diff
@@ -114,39 +114,39 @@ export const Footer: React.FC<FooterType> = (props) => {
...
             networkId={networkId}
-            label={'Allowance manager contract'}
+            label="Allowance manager contract"
           />
-        ) : null}
+        )}
       </ContractsWrapper>
       <VersionsWrapper>
-        {url.web && VERSION ? (
+        {url.web && VERSION && (
```
W3stside, line 135: noiiice

src/utils/contract.ts
```diff
@@ -13,10 +13,10 @@ export function getContractAddressFromNetworkInfo(
   network: OptionalNetwork,
   networkInfo: ContractNetworkInfo,
 ): string | null {
-  if (!network) {
+  if (!network || networkInfo[network]?.address) {
```
W3stside, line 16: mm i think this should just be: 
```suggestion
  if (!network || !networkInfo[network].address) {
```
or am i wrong? :thinking: now im confused with the coalescing

W3stside (approved):
just concerned about the utils/contracts line 16 and my noodle brain

src/utils/contract.ts
alfetopito: You are right, I forgot to negate the second part.
But maybe the coalescing should stay. Will try it out.

anxolin (approved):

Issue #255, opened 2021-03-05
Issue loader: Strange effect


anxolin:
If you load an order, periodically it checks for update

If you carefully watch the bottom of the page, you'll see a spinner appear and disappear creating a weird effect

alfetopito:
Ah yes. This was intentional

#design-pending

xD

---

In summary, that's an indicator that the loading is happening on the background.
But at the time didn't really stop to think about making it look good.

@biocom can probably suggest a way to pass the message without looking so ugly?

anxolin:
I suggest we can have a global flag for loading (global state), then we can represent that the app "is doing sth" anywhere we want. A swap UI might want to show it in the widget, and the explorer might want to show something in the footer for example.

I'm happy leaving this PR open if we just do a PR to kill that loading. We update periodically, but don't show a spinner.

biocom:
If you want I can come up with a proposal for this. As you perhaps are suggesting @anxolin we could have a universal loader/status display. Or something else.

alfetopito:
Disabling the loading for now https://github.com/gnosis/gp-ui/pull/281

PR #281, opened 2021-03-08
Do not display loading indicator when order is already loaded

alfetopito:
Part of #255 

Now the loading will only appear on first load

### DIFF
--- a/src/components/orders/OrderDetails/index.tsx
+++ b/src/components/orders/OrderDetails/index.tsx
@@ -66,7 +66,7 @@ export const OrderDetails: React.FC<Props> = (props) => {
         <p key={key}>{errors[key]}</p>
       ))}
       {/* TODO: create common loading indicator */}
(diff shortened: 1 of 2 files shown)


src/components/orders/OrderDetails/index.tsx
```diff
@@ -66,7 +66,7 @@ export const OrderDetails: React.FC<Props> = (props) => {
         <p key={key}>{errors[key]}</p>
       ))}
       {/* TODO: create common loading indicator */}
-      {isLoading && <FontAwesomeIcon icon={faSpinner} spin size="3x" />}
+      {isLoading && (!order?.buyToken || !order?.sellToken) && <FontAwesomeIcon icon={faSpinner} spin size="3x" />}
```
W3stside, line 69: maybe make this a `const` ? or does TS complain?

W3stside (approved):

PR #283, opened 2021-03-09
Home search progress.

biocom:
- Adds a search input on the homepage.
- Stores the input value (search query) in state.
- White spaces are trimmed onChange (input).
- useHistory to route the user to find the order.

Screen Shot 2021-03-09 at 18 09 26


biocom:
> > Second feedback would be, although it's great it doesn't have a button. I think a button on the side makes it more bullet-proof, right? How does it work in mobile. Probably it would war with enter key, but in a touch experience it feels like a button would help.
> > Maybe become irrelevant if we do the dropdown in the future
> > image
> 
> It has a button? https://github.com/gnosis/gp-ui/pull/283/files#diff-7d25be4074e4d7dcc7b58eb12e1663f5173dd9d1d4ccb4c37e2e4076c4679ed5R97

Ah I see now that on the preview (deploy) the SVG is not loading for some reason. Should look like the screenshot. Will look into it.

src/apps/explorer/components/common/Search/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+    fill: ${({ theme }): string => theme.grey};
+  }
+
+  &:hover {
+    opacity: 1;
+  }
+`
+
+export const Search: React.FC = () => {
+  const [query, setQuery] = useState('')
```
W3stside, line 82: we should make this entire thing a hook

src/apps/explorer/components/common/Search/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+        <SearchIcon src={searchImg} />
+      </button>
+      <input
+        autoComplete="off"
+        type="search"
+        name="query"
+        value={query}
+        onChange={(e): void => setQuery(e.target.value.trim())}
+        placeholder="Search by order ID"
+        aria-label="Search the GP explorer for orders, batches and transactions .r"
```
alfetopito, line 107: What's the `.r` at the end?

src/apps/explorer/components/common/Search/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+    &::placeholder {
+      color: inherit;
+      transition: color 0.2s ease-in-out;
+    }
+
+    &:focus::placeholder {
+      color: transparent;
+    }
+  }
```
alfetopito, line 46: I think this could be a separated styled component. Feels too big IMO.

alfetopito (approved):
Awesome!

Only missing the icon as Anxo pointed out

src/apps/explorer/components/common/Search/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+`
+
+export const Search: React.FC = () => {
+  const [query, setQuery] = useState('')
+  const history = useHistory()
+  const { path } = useRouteMatch()
+  const pathPrefix = path == '/' ? '' : path
+
+  const handleSubmit = useMemo(
```
W3stside, line 87: this should be `useCallback` `useMemo` returns `values` and here you're passing what seems like a function that returns a function to make it work

src/apps/explorer/components/common/Search/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+    <Wrapper onSubmit={handleSubmit}>
+      <button type="submit">
+        <SearchIcon src={searchImg} />
+      </button>
+      <input
+        autoComplete="off"
+        type="search"
+        name="query"
+        value={query}
+        onChange={(e): void => setQuery(e.target.value.trim())}
```
W3stside, line 105: this is imo a good time to use a memoised handler callback in the body that accepts `e` and does the logic inside the body with memoisation on the value as to not unnecessarily applying logic to the same input values

W3stside (changes requested):

src/apps/explorer/components/common/Search/index.tsx
biocom: @W3stside Thanks for the suggestion, will add it in this PR.

src/apps/explorer/components/common/Search/index.tsx
biocom: Typoo00.0..ooo

src/apps/explorer/components/common/Search/index.tsx
biocom: Agreed, will separate!

src/apps/explorer/components/common/Search/index.tsx
biocom: In this case the only logic applied is the `.trim()` function, which is also used to influence the visible input text (no spaces). Is that still needed? Or perhaps I misunderstand this.

src/apps/explorer/pages/Home/index.tsx
```diff
@@ -1,43 +1,43 @@
...
-    margin: 0 auto 20rem;
-    width: 100%;
-    max-width: 140rem;
-    padding: 0 1.6rem;
-
-    ${media.mediumDown} {
-      max-width: 94rem;
-    }
+  ${media.mediumDown} {
+    max-width: 94rem;
```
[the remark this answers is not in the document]
biocom: You are right. I'm thinking to bring this up in the parent page wrapper, so it applies to all pages.

src/hooks/useSearchSubmit.ts
```diff
@@ -0,0 +1,21 @@
+import { useCallback } from 'react'
```
biocom, line 1: This hook is intended to be used in multiple places:
- Homepage search
- Header search

src/apps/explorer/pages/Home/index.tsx
```diff
@@ -1,6 +1,5 @@
 import React from 'react'
 import styled from 'styled-components'
-import { media } from 'theme/styles/media'
```
biocom, line 3: Moved the max-width's to the App's parent wrapper.

src/apps/explorer/components/common/Search/index.tsx
```diff
@@ -1,99 +1,21 @@
...
-      e.preventDefault()
-      history.push(`${pathPrefix}/orders/${query}`)
-    },
-    [query, history, pathPrefix],
-  )
+  const handleSubmit = useSearchSubmit()
 
   return (
-    <Wrapper onSubmit={handleSubmit}>
+    <Wrapper
```
biocom, line 13: With @anxolin 's assistance, I moved this into a separate Hook, as per @W3stside 's suggestion.

src/apps/explorer/components/common/Search/index.tsx
W3stside: i would do this elsewhere (not this PR)

src/apps/explorer/components/common/Search/Search.styled.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    &::placeholder {
+      color: inherit;
+      transition: color 0.2s ease-in-out;
+    }
+
+    &:focus::placeholder {
+      color: transparent;
+    }
+  }
```
alfetopito, line 41: You said you would do another component. Changed your mind?

alfetopito (approved):

src/apps/explorer/components/common/Search/Search.styled.ts
biocom: Oh I might have misunderstood you. I thought you meant to separate the styles, but now I read you meant to create a specific styled component for the `input` element?

src/apps/explorer/components/common/Search/index.tsx
biocom: @W3stside in this case already in this PR. Will keep it in mind for future PR's.

src/apps/explorer/components/common/Search/Search.styled.ts
alfetopito: Yes, that's what I meant.
Only asked again because I thought you forgot.
No worries if you rather keep it as is.

biocom: I think your comment makes sense, especially in terms of readability. I will change it for this PR as it's a rather small change.

src/apps/explorer/components/common/Search/index.tsx
anxolin: I wouldn't memoize for saving the creation of the arrow function here. I would think this is fine

src/apps/explorer/pages/Home/index.tsx
```diff
@@ -1,43 +1,33 @@
...
-
-import homeTeaser from './home-teaser.png'
+import { Search } from 'apps/explorer/components/common/Search'
 
 const Wrapper = styled.div`
   display: flex;
   justify-content: center;
+  flex-flow: column wrap;
+  height: calc(100vh - 15rem);
```
anxolin, line 9: Should we have a page component, or some kind of container that do this? I mean, there's many places where we have the issue of the footer jumping to the top.

Also should it be min-height or height?

anxolin (approved):
@biocom Approved, I would encourage you to do a follow up issue or PR for making some basic storybook for the search component

src/apps/explorer/pages/Home/index.tsx
biocom: For the homepage doing a height, to have somewhat of a fixed height layout, where we keep in mind the footer. Many ways about it.

As for a page component: In my opinion, when you say the 'footer jumps' that is an effect of a page without content and/or flash of content. We therefore should solve that in individual cases: A normal page with at least a minimum amount of content never will have this problem.

On the order detail page the cause of the footer to perceivably 'jump' is because initially it is loading the order. We should address that by setting a minimum container height. Or even better, render the full table with all (minimum) columns but in a loading state for the values. 

Again, think we should address that separately.

alfetopito (approved):

src/apps/explorer/pages/Home/index.tsx
anxolin: I'm referring to #285

PR #290, opened 2021-03-12
Find order, even if you are not in the right network

anxolin:
So many times I found myself not finding an order, to later discover I was not in the correct network. 

I thought, this would happen quite some times, since we are jumping between networks, and an user arriving at the explorer would load mainnet by default, but he could potentially paste an ID from other network.

## Contains
The core of this is a recursive function `_getOrder`, that wraps the  `getOrder` API call. It will try to get the ID, if it fails, it will continue trying with a different network. 

Now the `useOrder` hooks return a new prop. An `errorOrderPresentInNetworkId`, that when returned, they signal that the order was not found because it's from another network. It is of type `Network`, so it also will tell us which network is the good one. 

Last but not list, the `OrderWidget` now checks for this error, and redirects the user to the right network. For that, it uses a new component that will redirect to the same path, but in the correct network.

## Test 
To try it out, use these IDs, and mess with it. Try to load the ids from the wrong network:

* mainnet: 0xf3f5e664b7337ff732b9c3e8d7d01a3394935e82a3fa466fe27ad8199b39c4494116d268a31493559dcf5eede2a83a00da9e0c8b604b79b6
* xdai: 0x203ab964d93ed04257dc6580f6b02f69a3e120104d9618efb28b952aaf00ae47424a46612794dbb8000194937834250dc723ffa5602a6c7c
* rinkeby: 0x88c5fd7b5537a30b71ab21af8e583e653734817da4ed37c4adbc0b20cd02b8b0424a46612794dbb8000194937834250dc723ffa5604b75e6


src/hooks/useOperatorOrder.ts
```diff
@@ -1,30 +1,72 @@
...
+  networkId: Network,
+  orderId: string,
+  networkIdSearchListRemaining: Network[],
+): Promise<GetOrderResult> {
+  const order = await getOrder({ networkId, orderId })
+
+  if (order || networkIdSearchListRemaining.length == 0) {
+    // We found the order in the right network
+    // or we have no more networks where to continue looking, so we return the "order" (can be null if it wasn't found)
+    return { order: order }
```
alfetopito, line 37: Nitpick:

```suggestion
    return { order }
```

src/hooks/useOperatorOrder.ts
```diff
@@ -1,30 +1,72 @@
...
+}
+
+async function _getOrder(
+  networkId: Network,
+  orderId: string,
+  networkIdSearchListRemaining: Network[],
+): Promise<GetOrderResult> {
+  const order = await getOrder({ networkId, orderId })
+
+  if (order || networkIdSearchListRemaining.length == 0) {
```
alfetopito, line 34: Nitpick:

```suggestion
  if (order || networkIdSearchListRemaining.length === 0) {
```

:shrug:

src/hooks/useOperatorOrder.ts
```diff
@@ -1,30 +1,72 @@
...
+    // We found the order in the right network
+    // or we have no more networks where to continue looking, so we return the "order" (can be null if it wasn't found)
+    return { order: order }
+  }
+
+  // If we didn't find the order in the current network, we look in different networks
+  const [nextNetworkId, ...remaindingNetworkIds] = networkIdSearchListRemaining
```
alfetopito, line 41: Typo

```suggestion
  const [nextNetworkId, ...remainingNetworkIds] = networkIdSearchListRemaining
```

src/hooks/useOperatorOrder.ts
```diff
@@ -1,30 +1,72 @@
...
+  console.log('is in different network', isOrderInDifferentNetwork)
+  if (isOrderInDifferentNetwork) {
+    // If the order exist in the other network
+    return {
+      order: null,
+      errorOrderPresentInNetworkId: nextNetworkId,
+    }
+  } else {
+    // Keep looking in other networks
+    return _getOrder(nextNetworkId, orderId, remaindingNetworkIds)
```
alfetopito, line 57: Typo

```suggestion
    return _getOrder(nextNetworkId, orderId, remainingNetworkIds)
```

src/hooks/useOperatorOrder.ts
```diff
@@ -33,10 +75,18 @@ export function useOrder(orderId: string, updateInterval = 0): UseOrderResult {
       setError('')
 
       try {
-        const rawOrder = await getOrder({ networkId, orderId })
+        const { order: rawOrder, errorOrderPresentInNetworkId: errorOrderPresentInNetworkIdRaw } = await _getOrder(
+          networkId,
+          orderId,
+          NETWORK_ID_SEARCH_LIST.filter((network) => network != networkId),
```
alfetopito, line 81: I think this filtering belongs in `_getOrder`, not on the caller.
And this parameter be optional, defaulting to all networks.

src/hooks/useOperatorOrder.ts
```diff
@@ -1,30 +1,72 @@
...
+  order: RawOrder | null
+  errorOrderPresentInNetworkId?: Network
+}
+
+async function _getOrder(
+  networkId: Network,
+  orderId: string,
+  networkIdSearchListRemaining: Network[],
+): Promise<GetOrderResult> {
+  const order = await getOrder({ networkId, orderId })
```
alfetopito, line 32: I'm wondering, from the UX point of view, wouldn't it be better if we fetch all networks at once instead of trying sequentially?

Or only after the initial network id query fails.

Not sure, just thinking out loud.

alfetopito:
Also, browser's back action doesn't work after the redirect.

src/hooks/useOperatorOrder.ts
anxolin: I started doing like that for some reason. And I changed it to this, but I think u r right, I'll change it!

src/hooks/useOperatorOrder.ts
anxolin: I saw it's pretty fast. So I just felt it was fine. I can happily change this if ends up being not performant or something. but is a good point. Being sequential also do less calls. I look in order of relevance (mainnet, xdai, rinkeby)

anxolin:
> Also, browser's back action doesn't work after the redirect.

yep, saw that. Fixed!

biocom:
@anxolin Worked smooth. One thing to add: Probably would be nice to have a general purpose notification/toast, which very subtly tells you that a different chain (from mainnet) was loaded. Nothing obtrusive and nothing where user interactivity is required by the user to 'consent'. But just to inform the user. 

Right now you'd tel by seeing the Xdai/Rinkeby label at the logo and/or just by knowing it's an order from a different chain. But likely good to address this in future PR's.

biocom (approved):

anxolin:
@W3stside please, still do the review. I'm merging to be able to freeze version

src/hooks/useOperatorOrder.ts
```diff
@@ -1,30 +1,74 @@
...
+import { Order, getOrder, RawOrder } from 'api/operator'
 
 import { transformOrder } from 'utils'
 
 import { useNetworkId } from 'state/network'
 
 import { useMultipleErc20 } from './useErc20'
+import { Network } from 'types'
+
+const NETWORK_ID_SEARCH_LIST = [Network.Mainnet, Network.xDAI, Network.Rinkeby]
```
W3stside, line 12: not sure how we feel about this but it is dynamic:
```ts
const NETWORK_VALUES = Object.values(Network)
const NETWORK_ID_SEARCH_LIST = NETWORK_VALUES.slice(NETWORK_VALUES.length / 2)
// ["1", "3", "4", "42", "100"]
```

src/hooks/useOperatorOrder.ts
```diff
@@ -1,30 +1,74 @@
...
+    // so we return the "order" (can be null if it wasn't found in any network)
+    return { order }
+  }
+
+  // If we didn't find the order in the current network, we look in different networks
+  const [nextNetworkId, ...remainingNetworkIds] = networkIdSearchListRemaining.filter((network) => network != networkId)
+
+  // Try to get the oder in another network (to see if the ID is OK, but the network not)
+  const isOrderInDifferentNetwork = await getOrder({ networkId: nextNetworkId, orderId }).then(
+    (order) => order !== null,
```
W3stside, line 47: are `{}` orders accepted? is falsy all bad? or just `null`?
could be written as `.then(Boolean)`

src/hooks/useOperatorOrder.ts
```diff
@@ -33,10 +77,17 @@ export function useOrder(orderId: string, updateInterval = 0): UseOrderResult {
       setError('')
 
       try {
-        const rawOrder = await getOrder({ networkId, orderId })
+        const { order: rawOrder, errorOrderPresentInNetworkId: errorOrderPresentInNetworkIdRaw } = await _getOrder(
+          networkId,
+          orderId,
+        )
+        console.log({ rawOrder, errorOrderPresentInNetworkIdRaw })
```
W3stside, line 84: you should remove this log or make it more descriptiev

src/state/network/updater.tsx
```diff
@@ -9,15 +9,40 @@ import { useNetworkId } from './hooks'
...
-      return Network.xDAI
-    default:
-      return Network.Mainnet
+const MAINNET_PREFIX = ''
+const NETWORK_PREFIXES_RAW: [Network, string][] = [
+  [Network.Mainnet, ''],
+  [Network.xDAI, 'xdai'],
+  [Network.Rinkeby, 'rinkeby'],
+]
+const PREFIX_BY_NETWORK_ID: Map<Network, string> = new Map(NETWORK_PREFIXES_RAW)
```
W3stside, line 18: why not just use the `enum` here instead of all these `Maps`? 

```ts
enum ChainId {
    MAINNET = 1,
    ROPSTEN = 3,
    RINKEBY = 4,
    GÖRLI = 5,
    KOVAN = 42,
    XDAI = 100
}

const network = 1
const networkName = "MAINNET"

console.log(ChainId[network])
// MAINNET
console.log(ChainId[networkName])
// 1
```

W3stside:
approved with comments

PR #322, opened 2021-03-23
Move dimensions to const

anxolin:
Move the constants to the right place (constant file instead of theme 🙈)

### DIFF
--- a/src/const.ts
+++ b/src/const.ts
@@ -224,8 +224,3 @@ export const DISABLED_TOKEN_MAPS = Object.keys(disabledTokens).reduce<DisabledTo
     [Network.xDAI]: {},
   },
 )
-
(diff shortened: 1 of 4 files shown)


src/const.ts
```diff
@@ -224,3 +224,8 @@ export const DISABLED_TOKEN_MAPS = Object.keys(disabledTokens).reduce<DisabledTo
     [Network.xDAI]: {},
   },
 )
+
+export enum AnalyticsDimension {
```
alfetopito, line 228: Since this is a const, shouldn't it be all caps?

alfetopito (approved):

src/const.ts
anxolin: actually is a enum so i moved it to types

PR #392, opened 2021-04-22
291/header link issues

### LINKED ISSUE #291: Issue with link in the header
I think the header has some code from the trading ui

image

Also, the link should be not only for the logo. If we click GP Explorer, I think it should get to the home. 

Lastly, the link to the home (and any link within the app, really, should be within the same network). Right now, if you are in rinkeby, and click home, you go to mainnet. 

cc/ @alfetopito

alfetopito:
# Edit 2021/04/26:

- Rename header to `Gnosis Protocol`
- Made only the logo spin

screenshot_2021-04-26_17-02-26

# Summary

Closes #291 

- Makes the whole header logo + label clickable (and spinnable)
- Link now keeps selected network

@biocom probably styling fixes needed :sweat_smile: 
Not sure about the whole logo spinning thing, maybe we want to spin only the image =P

# Testing

1. Open app, search order ids from different network:
    - mainnet: 0x3da38a8e6d9725b6d1f1d1619004c5d2cd4cd1ff5db985ee9965abd9bfe22d097b2e78d4dfaaba045a167a70da285e30e8fca196606dbb23
    - rinkeby: 0x0f72377ed644b8bed0e1b8540904dc5b2308c17db3c0723025a60c1c1dde551a5b0abe214ab7875562adee331deff0fe1912fe42607df52d
    - xdai: 0xc64dae2bde7f4538030e056aace4312f8208e7144fd44aa3d018e7c4a574294804a66cbba0485d7b21af836f52b711401300fddb6066027c
2. For every order id:
    1. Search on home using the given id
    - [ ] Should redirect the appropriated network and load the order
    2. Click on the logo
    - [ ] Should load home in the same network

- [ ] Verify the whole logo (image and label) are clickable

### DIFF
--- a/src/apps/explorer/layout/Header.tsx
+++ b/src/apps/explorer/layout/Header.tsx
@@ -43,13 +43,13 @@ export const Header: React.FC = () => {
     return null
   }
 
-  const network = networkId !== 1 ? getNetworkFromId(networkId) : null
+  const network = networkId !== 1 ? getNetworkFromId(networkId).toLowerCase() : null
 
   return (
-    <GenericHeader>
+    <GenericHeader logoAlt="GP Explorer homepage" linkTo={`/${network || ''}`}>
       <Navigation>
         <Logo>GP Explorer</Logo>
-        {network && <NetworkLabel className={network.toLowerCase()}>{network}</NetworkLabel>}
+        {network && <NetworkLabel className={network}>{network}</NetworkLabel>}
         {/*      
         <li>
           <Link to="/">Batches</Link>
--- a/src/apps/explorer/layout/Header.tsx
+++ b/src/apps/explorer/layout/Header.tsx
@@ -13,6 +13,7 @@ const Logo = styled.span`
   font-size: 1.8rem;
   line-height: 1;
   font-weight: ${({ theme }): string => theme.fontBlack};
+  white-space: nowrap;
 `
 
 const NetworkLabel = styled.span`
@@ -46,9 +47,8 @@ export const Header: React.FC = () => {
   const network = networkId !== 1 ? getNetworkFromId(networkId).toLowerCase() : null
 
   return (
-    <GenericHeader logoAlt="GP Explorer homepage" linkTo={`/${network || ''}`}>
+    <GenericHeader logoAlt="GP Explorer homepage" linkTo={`/${network || ''}`} label={<Logo>GP Explorer</Logo>}>
       <Navigation>
-        <Logo>GP Explorer</Logo>
         {network && <NetworkLabel className={network}>{network}</NetworkLabel>}
         {/*      
         <li>
--- a/src/apps/explorer/layout/Header.tsx
+++ b/src/apps/explorer/layout/Header.tsx
@@ -47,7 +47,7 @@ export const Header: React.FC = () => {
   const network = networkId !== 1 ? getNetworkFromId(networkId).toLowerCase() : null
 
   return (
-    <GenericHeader logoAlt="GP Explorer homepage" linkTo={`/${network || ''}`} label={<Logo>GP Explorer</Logo>}>
+    <GenericHeader logoAlt="Gnosis Protocol" linkTo={`/${network || ''}`} label={<Logo>Gnosis Protocol</Logo>}>
       <Navigation>
         {network && <NetworkLabel className={network}>{network}</NetworkLabel>}
         {/*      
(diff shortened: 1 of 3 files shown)


anxolin (commented):
Great PR, not approving just for a nitpick :) 

I don't like the effect on the logo. I like that now you get the effect when you hover on the text too, but can we just make the effect on the OWL logo?

Also, I suggest renaming to:
image

biocom:
> Great PR, not approving just for a nitpick :)
> 
> I don't like the effect on the logo. I like that now you get the effect when you hover on the text too, but can we just make the effect on the OWL logo?
> 
> Also, I suggest renaming to:
> image

- Agreed, keep the effect only to the OWL icon. 
- Would be in favor of a shorter title -> GP Explorer. Much more recognizable.

anxolin:
> Would be in favor of a shorter title -> GP Explorer. Much more recognizable.

Yep, I like it short, but we received the feeback people doesn't know what is GP.
Maybe we can do a compromise, we show GP Explorer, and then a subtitle below that includes Gnosis Protocol somehow?
Or we show Gnosis Protocol, and then in another color and text size "Explorer"

@annamsgeorge any thoughts?

biocom:
@anxolin 
- Could you elaborate on what user(s) type(s) you got the feedback from? 
- We could do something like this:
Screen Shot 2021-04-23 at 17 05 20
But actually recommend against it, as it feels like cluttering the header.
- The browser URL bar clearly shows `https://gnosis-protocol.io/` so still questioning the need to add it, without knowing anything about the background of this feedback.

anxolin:
If you are not convinced we can wait for more feedback, or ask arounnd:

Marce: 
image

Lukas (I think he is critique is not to the acronym, but somehow related):
image

annamsgeorge:
I agree with Anxo - it's not clear to the user what GP stands for. It's better if we spell it out for the beginning. Sometime in the future when it's more clear to users we can rebrand it to GP, but for the beginning it's important to educate users on the meaning of GP.

annamsgeorge:
PS: I like Michel's proposal 
Screenshot 2021-04-23 at 11 57 20

biocom:
> If you are not convinced we can wait for more feedback, or ask arounnd:
> 
> Marce:
> image
> 
> Lukas (I think he is critique is not to the acronym, but somehow related):
> image

Thanks for sharing @anxolin 

My understanding is, that this is feedback for our swapping UI. Obviously that's somewhat related to the explorer, but that would suggest we'd call it 'Gnosis Protocol' everywhere, including in the swapping UI?

Otherwise, I'm good with @annamsgeorge 's suggestion to use this for the time being. I'll address this in a PR. 

At the same time, I don't see this solving for the above referenced feedback, given that's related to the swapping UI. Curious on your thoughts.

anxolin:
Yep, is from the UI, is not 100% 1-1 applicable here, it's just that, at least for now GP don't mean too much for the users, so it's nice to be a bit more explicit while we build brand.

> At the same time, I don't see this solving for the above referenced feedback, given that's related to the swapping UI. Curious on your thoughts.

Agree, it doesn't fully solve it. Although I do think is an improvement. Your proposal gives more context:

--------
**GP Explorer**
Gnosis Protocol
--------

I like it, although I would probably prefer something like:

--------
**Gnosis Protocol**
Explorer
--------

biocom:
> Yep, is from the UI, is not 100% 1-1 applicable here, it's just that, at least for now GP don't mean too much for the users, so it's nice to be a bit more explicit while we build brand.
> 
> > At the same time, I don't see this solving for the above referenced feedback, given that's related to the swapping UI. Curious on your thoughts.
> 
> Agree, it doesn't fully solve it. Although I do think is an improvement. Your proposal gives more context:
> 
> .........
> **GP Explorer**
> Gnosis Protocol
> .........
> 
> I like it, although I would probably prefer something like:
> 
> .........
> **Gnosis Protocol**
> Explorer
> .........

How about calling it just *Gnosis Protocol* without the 'explorer' part? I'd envision it being evident to most users what gnosis-protocol.io is (coming in from a submitted/confirmed tx link on the Swap UI).

anxolin:
> How about calling it just Gnosis Protocol without the 'explorer' part? I'd envision it being evident to most users what gnosis-protocol.io is (coming in from a submitted/confirmed tx link on the Swap UI).

sounds good to me 👌

anxolin (approved):
Great!

Approving and merging for the release. 
As a uber-mega-nitpick, my comment of flipping only the gno logo and not the text referred to flipping it also if you hover the text. So the animation only for the logo, but the active area is all the logo. Anyways, is not important, merging!

alfetopito:
> Great!
> 
> Approving and merging for the release.
> As a uber-mega-nitpick, my comment of flipping only the gno logo and not the text referred to flipping it also if you hover the text. So the animation only for the logo, but the active area is all the logo. Anyways, is not important, merging!

Oh sorry, did not get that

PR #406, opened 2021-04-26
Deploy tags automatically

anxolin:
**Description**
Prepare PRODUCTION deploy automatically on every tag.

**Why**
Because before we needed to tag, and then create a new release. We want to prepare for deployment every tag.

Also, there were some small errors explained in the details

**Details**
Now it gives better output. It prints the current version (so we also make sure the env var is correctly set)

Also, it will inform about the passed parameters. If they are not available:
image

Additionally, now returns an error in that case. Before this PR it was just printing the error, and showing success (ends with a 0 error code). So now, if any param is missing, now it returns the `100` error code.

The error message, now is printed in stderr, instead of stdout (more correct, but also GA understand it and display it better)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,140 +15,140 @@ env:
   PR_NUMBER: ${{ github.event.number }}
 
 jobs:
-  # setup:
-  #   name: Setup
-  #   runs-on: ubuntu-latest
+  setup:
+    name: Setup
+    runs-on: ubuntu-latest
 
-  #   steps:
-  #     - name: Checkout code
-  #       uses: actions/checkout@v2
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
 
-  #     - name: Set up node
-  #       uses: actions/setup-node@v2
-  #       with:
-  #         node-version: 12
+      - name: Set up node
+        uses: actions/setup-node@v2
+        with:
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@gnosis.pm/gp-v1",
-  "version": "2.1.0-rc.0",
+  "version": "2.1.0-rc.17",
```
alfetopito, line 3: 17? LoL

anxolin: Yes!

I was commenting most of the cy.yml and making looots of tests :)

davidalbela (approved):

.github/workflows/ci.yml
```diff
@@ -153,10 +152,13 @@ jobs:
...
   deploy:
     name: Deploy
-    needs: [build, storybook]
+    # needs: [build, storybook]
     runs-on: ubuntu-latest
 
     steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
```
alfetopito, line 161: Don't care about the storybook, but no longer depends on `build`?

I see you added the code checkout step, but why? Need to have the source in order to push stuff?

anxolin: yes, that was a test!
forgot to uncomment

PR #553, opened 2021-07-29
Fix lint issues

anxolin:
Solves some lint issues that I assume comes from updates in prettier and eslint done by dependabot automatically:
```
✖ 176 problems (176 errors, 0 warnings)
  176 errors and 0 warnings potentially fixable with the `--fix` option.
ERROR: "lint:js" exited with 1.
husky > pre-commit hook failed (add --no-verify to bypass)
```

Surprising that the bot managed to commit without husky running the linter, plus I'm also surprise the CI didn't catch this. Maybe is good to take a look why this happened


henrypalacios (approved):
With these changes I was able to commit without previous 176 errors.

alfetopito (approved):

PR #560, opened 2021-08-02
Create card component

stackseeder:
# Summary

Closes https://github.com/gnosis/cowswap/issues/1042

Add a Card and CardRow components

  # To Test
1. Visit https://pr560--gpui.review.gnosisdev.com/storybook/
- [ ] Under Common should be a Card section
- [ ] Under Common should be a CardRow section

### DIFF
--- a/src/components/common/Card/index.tsx
+++ b/src/components/common/Card/index.tsx
@@ -28,6 +28,9 @@ const CardComponent = styled(DefaultCard)`
 // CARD CONTENT STYLES
 const CardContent = styled.div`
   flex: 1;
+  display: flex;
+  justify-content: center;
+  align-items: center;
   font-size: 15px;
   padding: 16px;
   line-height: normal;
--- a/src/components/common/Card/index.tsx
+++ b/src/components/common/Card/index.tsx
@@ -1,11 +1,10 @@
 import React from 'react'
-import styled, { css } from 'styled-components'
-import { variants } from 'styled-theming'
+import styled from 'styled-components'
 
 import { COLOURS } from 'styles'
 import { Theme } from 'theme'
 
-const { white, whiteDark, fadedGreyishWhite, blackLight } = COLOURS
+const { white, fadedGreyishWhite, blackLight } = COLOURS
 
 const DefaultCard = styled.div`
   height: inherit;
@@ -16,38 +15,14 @@ const DefaultCard = styled.div`
   box-shadow: 0 10px 15px -3px rgb(0 0 0 / 7%), 0 4px 6px -2px rgb(0 0 0 / 5%);
   margin: 10px;
 `
-// Create our variated Card Theme
-// 'variant' refers to a prop on button
-// <CardBase variant="danger" />
-export const CardTheme = variants('mode', 'variant', {
-  default: {
-    [Theme.LIGHT]: css`
-      background: ${white};
-      color: ${blackLight};
-      border: 1px solid ${white};
-    `,
-    [Theme.DARK]: css`
-      background: ${fadedGreyishWhite};
-      color: ${white};
-      border: 1px solid ${whiteDark};
-    `,
-  },
-  get primary() {
-    return this.default
-  },
-})
 
-const StyledCard = styled(DefaultCard)`
-  /* Fold in theme css above */
-  ${CardTheme}
-`
-
-const CardComponent = styled(StyledCard)<{ outline?: boolean }>`
+const CardComponent = styled(DefaultCard)`
   display: flex;
   flex-direction: column;
   border-top-right-radius: 6px;
   border-top-left-radius: 6px;
-  border-width: ${({ outline }): string => (outline ? '1px' : '0px')};
+  background: ${({ theme }): string => (theme.mode == Theme.DARK ? fadedGreyishWhite : white)};
+  color: ${({ theme }): string => (theme.mode == Theme.DARK ? white : blackLight)};
 `
 
 // CARD CONTENT STYLES
@@ -61,7 +36,6 @@ const CardContent = styled.div`
 `
 
 export interface CardBaseProps {
-  outline?: boolean
   children?: React.ReactElement
 }
 
@@ -70,13 +44,9 @@ export interface CardBaseProps {
  *
  * An extensible content container.
  */
-export const Card: React.FC<
-  CardBaseProps & {
-    outline?: boolean
-  }
-> = ({ children, outline, ...rest }) => {
+export const Card: React.FC<CardBaseProps> = ({ children, ...rest }) => {
   return (
-    <CardComponent {...rest} variant={'default'} outline={outline || false}>
+    <CardComponent {...rest}>
       <CardContent>{children}</CardContent>
     </CardComponent>
   )
--- a/src/components/common/Card/index.tsx
+++ b/src/components/common/Card/index.tsx
@@ -1,31 +1,18 @@
-import React, { useEffect, useState } from 'react'
+import React from 'react'
 import styled, { css } from 'styled-components'
 import { variants } from 'styled-theming'
 
 import { COLOURS } from 'styles'
 import { Theme } from 'theme'
 
-const {
-  white,
-  whiteDark,
-  blue,
-  successLight,
-  successDark,
-  warningLight,
-  warningDark,
-  dangerLight,
-  dangerDark,
-  bgLight,
-  bgDark,
-  blackLight,
-} = COLOURS
+const { white, whiteDark, bgDark, blackLight } = COLOURS
 
 const DefaultCard = styled.div`
   height: inherit;
   min-width: 200px;
   min-height: 100px;
   background-color: #f5f5f5;
-  border-radius: 10px;
+  border-radius: 6px;
   box-shadow: 0 10px 15px -3px rgb(0 0 0 / 7%), 0 4px 6px -2px rgb(0 0 0 / 5%);
   margin: 10px;
 `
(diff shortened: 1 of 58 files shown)


ramirotw (commented):
notes:
- do we need card states? (info, success, warning, danger). Based on the wireframes I couldn't find a use case for it
- does the card needs either a header or a footer?

@alfetopito wdyt?

src/components/common/Card/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+  flex-direction: column;
+  border-top-right-radius: 10px;
+  border-top-left-radius: 10px;
+  border-width: ${({ outline }): string => (outline ? '1px' : '0px')};
+`
+
+// CARD CONTENT STYLES
+const CardContent = styled.div<{ $hideHeader: boolean }>`
+  flex: 1;
+  border-top-right-radius: ${({ $hideHeader }): string => ($hideHeader ? '10px' : '0px')};
```
alongoni, line 128: border-radius should be 6px.

src/components/common/CardRow/CardRow.stories.tsx
```diff
@@ -0,0 +1,101 @@
...
+  }
+  span {
+    font-weight: bold;
+    font-size: 12px;
+    margin: 2px 1px 2px 0px;
+    &.red {
+      color: red !important;
+    }
+    &.green {
+      color: #06b506 !important;
```
alongoni, line 36: #00D897 (use this green)

src/components/common/CardRow/CardRow.stories.tsx
```diff
@@ -0,0 +1,101 @@
...
+  }
+  h3 {
+    margin: 2px 3px -1px 3px;
+  }
+  span {
+    font-weight: bold;
+    font-size: 12px;
+    margin: 2px 1px 2px 0px;
+    &.red {
+      color: red !important;
```
alongoni, line 33: #F1356E (use this red)

src/components/common/Card/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+`
+// Create our variated Card Theme
+// 'variant' refers to a prop on button
+// <CardBase variant="danger" />
+export const CardTheme = variants('mode', 'variant', {
+  default: {
+    [Theme.LIGHT]: css`
+      background: ${white};
+      color: ${blackLight};
+      border-top: 3px solid ${white} !important;
```
alongoni, line 40: Please, remove 3px border on top.

src/components/common/Card/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+      border: 1px solid ${white};
+    `,
+    [Theme.DARK]: css`
+      background: ${bgDark};
+      color: ${white};
+      border-top: 3px solid ${whiteDark} !important;
```
alongoni, line 46: Please, remove 3px border on top.

src/components/common/Card/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+    `,
+  },
+  get primary() {
+    return this.default
+  },
+  info: {
+    [Theme.LIGHT]: css`
+      color: ${blackLight};
+      background: ${white};
+      border-top: 3px solid ${blue} !important;
```
alongoni, line 57: Please, remove 3px border on top.

src/components/common/Card/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+      border: 1px solid ${blue};
+    `,
+    [Theme.DARK]: css`
+      color: ${white};
+      background: ${bgDark};
+      border-top: 3px solid ${blue} !important;
```
alongoni, line 63: Please, remove 3px border on top.

alongoni (changes requested):
Looks great! I've made some styles suggestions.

src/components/common/CardRow/CardRow.stories.tsx
alfetopito: Or better yet, take it from the theme, see how it's done in other components https://github.com/gnosis/gp-ui/blob/4da5595b204f6dc2b59bf4bb61c809cb37221ece/src/components/orders/OrderSurplusDisplay/index.tsx#L21

alfetopito:
> notes:
> 
>     * do we need card states? (info, success, warning, danger). Based on the wireframes I couldn't find a use case for it
> 
>     * does the card needs either a header or a footer?
> 
> 
> @alfetopito wdyt?

Agree with Ramiro. No need for states nor header/footer.

Based on the discussion last week about this (sorry have not written it down in the task), the suggestion is to keep the Card and CardRow components generic and simple with the outer style fixed, but the inner contents style defined by the parent component.

So you can split the work in 2 PRs
1. The first (this one) defining the Card and CardRow components, where the card takes in the content as a prop. Or even as children, no strong preference here.
2. A follow up where you add a component that implements the contents of the styles (3 rows, 2 variations) as per latest design:
screenshot_2021-08-03_10-30-07

alongoni:
I forgot one small detail:
The background color of the cards should be solid (not a gradient). Thanks!

src/theme/styles/colours.ts
```diff
@@ -60,6 +61,8 @@ export const BASE_COLOURS = {
   red1: '#FF305B',
   red2: '#FF6871',
   red3: '#F82D3A',
+  red4: '#F13E6E',
+  green: '#41C29B',
```
alfetopito, line 65: If you are adding `red4`, do the same for `green` and add `green4`.

But actually, why not to use the existing reds and greens?
From the designs the colors should be the same as the ones already in place.

src/toastify.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
alfetopito, line 1: And, this is the file causing problems.
I believe there were some dependabot updates failing because of this. 

Thanks!

alfetopito (commented):
Partial review, couldn't finish today, will continue tomorrow

src/components/common/Card/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+  border-width: ${({ outline }): string => (outline ? '1px' : '0px')};
+`
+
+// CARD CONTENT STYLES
+const CardContent = styled.div`
+  flex: 1;
+  border-top-right-radius: 6px;
+  border-top-left-radius: 6px;
+  font-size: 15px;
+  padding: 5px 10px 5px 10px;
```
alongoni, line 59: Maybe we can add more padding here: 
padding: 16px;

src/components/common/Card/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+}
+
+/**
+ * Card component.
+ *
+ * An extensible content container.
+ */
+export const Card: React.FC<
+  CardBaseProps & {
+    outline?: boolean
```
alfetopito, line 75: It seems to me that this is not changing anything in the type, is it?

So you don't really need this part `& { outline?: boolean }`

src/components/common/Card/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+  }
+> = ({ children, outline, ...rest }) => {
+  return (
+    <CardComponent {...rest} variant={'default'} outline={outline || false}>
```
alfetopito, line 79: I don't think this component needs a `variant`

alfetopito: Actually, maybe not even the `outline`.
We could have a highligh onHover, but that can be done in CSS, no need for the extra prop

src/components/common/CardRow/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * CardRow component.
+ *
+ * Place cards side-by-side
+ */
+export const CardRow: React.FC<
+  CardRowProps & {
+    overflow?: boolean
+  }
+> = ({ children, overflow }) => {
```
alfetopito, line 21: You don't need this here either
```suggestion
export const CardRow: React.FC<CardRowProps> = ({ children, overflow }) => {
```

alfetopito (changes requested):

src/theme/styles/colours.ts
stackseeder: @alfetopito I've checked the colors, but no existing one was matching to the design. So I've added new one.

alfetopito: I see. No worries, just pick of of the existing colors, I don't think we need to introduce new green/red for this component.

alfetopito: cc @alongoni

src/components/common/CardRow/CardRow.stories.tsx
```diff
@@ -0,0 +1,101 @@
...
+  }
+  span {
+    font-weight: bold;
+    font-size: 12px;
+    margin: 2px 1px 2px 0px;
+    &.red {
+      color: ${({ theme }): string => theme.red1} !important;
+    }
+    &.green {
+      color: ${({ theme }): string => theme.green} !important;
```
alfetopito, line 36: It's just for a storybook, pick one existing green instead of adding a new one

alfetopito (commented):
Almost there, just one small comment regarding one storybook `green` color

alfetopito (approved):
:+1:

src/components/common/CardRow/CardRow.stories.tsx
```diff
@@ -0,0 +1,101 @@
...
+  decorators: [GlobalStyles, ThemeToggler],
+} as Meta
+
+const CardBody = styled.div`
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  p {
+    margin-bottom: 0px;
```
alongoni, line 23: In order to center the content we can set this margin in 0: (`margin: 0` instead of `margin-botton:0px`)
image

Then we need to add 
    `display: flex;
    justify-content: center;
    align-items: center;`
image
****

src/components/common/CardRow/CardRow.stories.tsx
```diff
@@ -0,0 +1,101 @@
...
+            <span className="green">+1.03%</span>
+          </div>
+        </CardBody>
+      </Card>
+      <Card>
+        <CardBody>
+          <p>24H Voulme</p>
+          <div>
+            <h3>194</h3>
+            <span className="red">-3.45%</span>
```
alongoni, line 74: We can try to add a new story with 3 rows:
image

stackseeder: @alongoni how about to implement detailed card and variants in separate PR?
Otherwise this PR could be bigger in my opinion.

alfetopito: That's a great idea @stackseeder 

We can focus on nailing the styles in another PR.

src/toastify.ts
```diff
@@ -23,8 +22,16 @@ type ToastPromised = {
...
+      (): ToastContent =>
+      (
+        ...args: [
+          (((((ToastContent & ToastId) & (ToastId | undefined)) & ToastId) &
+            ((count?: number | undefined, containerId?: string | number | undefined) => void)) &
+            ToastId) &
+            (ToastContainerProps | undefined),
+          ToastOptions | undefined,
+        ]
+      ): Promise<string | number | boolean | void> =>
```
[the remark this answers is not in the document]
stackseeder: @ramirotw I've reverted the change in the toastify.ts

PR #561, opened 2021-08-03
Create / Stylize Explorer-UserDetails Tab component

henrypalacios:
closes: #1043

- Create a new component that uses the [Tabs](https://github.com/gnosis/gp-ui/blob/styling-tab-1043/src/components/common/Tabs/Tabs.tsx#L62) as a Base and style it. **(path taken)**
    - [x] Use theme Provider variables.
    - [x]  Ability to use some variants of the [thickness](https://github.com/gnosis/gp-ui/blob/styling-tab-1043/src/components/common/Tabs/TabItem.tsx#L41) of the tab indicator
    - [x] Add an extra content to Tabs.

### DIFF
--- a/src/apps/explorer/components/common/ExplorerTabs/ExplorerTab.tsx
+++ b/src/apps/explorer/components/common/ExplorerTabs/ExplorerTab.tsx
@@ -3,6 +3,7 @@ import styled from 'styled-components'
 
 import Tabs, { getTabTheme, Props as TabsProps, IndicatorTabSize } from 'components/common/Tabs/Tabs'
 import { DARK_COLOURS } from 'theme'
+import { MEDIA } from 'const'
 
 const StyledTabs = styled.div`
   display: flex;
@@ -32,6 +33,14 @@ const StyledTabs = styled.div`
 
   > div > div.tab-content {
     padding: 20px 16px;
+  }
+
+  .tab-extra-content {
+    width: 100%;
+
+    @media ${MEDIA.mobile} {
+      display: none; /* for now we can hide the extra-content on mobiles */
+    }
   }
 `
 const tabCustomThemeConfig = getTabTheme({
--- a/src/apps/explorer/components/common/ExplorerTabs/ExplorerTab.tsx
+++ b/src/apps/explorer/components/common/ExplorerTabs/ExplorerTab.tsx
@@ -1,0 +1,59 @@
+import React from 'react'
+import styled from 'styled-components'
+
+import Tabs, { getTabTheme, Props as TabsProps } from 'components/common/Tabs/Tabs'
+import { DARK_COLOURS } from 'theme'
+
+const StyledTabs = styled.div`
+  display: flex;
+  width: 100%;
+  padding: 0;
+  border: 1px solid rgba(151, 151, 184, 0.3);
+  border-radius: 4px;
+
+  > div > div.tablist {
+    justify-content: flex-start;
+    border-bottom: 1px solid rgba(151, 151, 184, 0.3);
+    box-sizing: border-box;
+  }
+
+  > div > div.tablist > button {
+    flex: 0 0 auto;
+    min-width: 96px;
+    padding: 12px 0.8rem;
+    line-height: 2;
+    height: auto;
+    font-family: var(--font-default);
+  }
+
+  > div > div:last-of-type {
+    height: 100%;
+  }
+
+  > div > div.tab-content {
+    padding: 20px 16px;
+  }
+`
+const tabCustomThemeConfig = getTabTheme({
+  activeBg: 'var(--color-transparent)',
+  activeBgAlt: 'initial',
+  inactiveBg: 'var(--color-transparent)',
+  activeText: DARK_COLOURS.textPrimary1,
+  inactiveText: 'var(--color-text-secondary2)',
+  activeBorder: DARK_COLOURS.orange,
+  inactiveBorder: 'none',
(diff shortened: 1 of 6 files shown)


alfetopito (commented):
Looking at the result with the custom theme, doesn't look too far off.

What is the problem you are having?

I have not been able to sync with @biocom yet, but maybe he has any input?

henrypalacios:
@alfetopito I have create another [component](https://github.com/gnosis/gp-ui/blob/styling-tab-1043/src/apps/explorer/components/common/ExplorerTabs/ExplorerTab.tsx#L51) using Tabs as a base.

My concern is if I have to edit something to achieve in the future to use the space after the tabs for other components **(as well as in design)**

ExplorerTabs

biocom:
@henrypalacios I think given we inherit the Tab component here, I think we need to edit the original component. Perhaps by having specifying different 'Tab' types? E.g. normal vs. special purpose. Remind me what was the design we are trying to implement?

biocom:
Screen Shot 2021-08-04 at 12 40 52

If it's to implement this ^, would the Tab component need to be edited at all? My thinking (correct me if I got the intent wrong):

```
<>
  <Tabs />
  <Search />
  <Pagination />
  <Filters />
</>
```

henrypalacios:
Thanks  for you point of view @biocom, Looking at your last comment with the drawing of the boxes, you have hit the point. 

Following this:
```
<>
  <Tabs />
  <Search />
  <Pagination />
  <Filters />
</>
```
I could place the other components next to the tabs (green box), but I'm not sure if I can get the `div` with the content tab to stay underneath it at 100%. I will give it a try and keep you posted.

alongoni:
I think it will work fine! I've made a quick test and looks good:
image
image

src/components/common/Tabs/Tabs.tsx
```diff
@@ -45,7 +45,6 @@ const Wrapper = styled.div`
     width: 100%;
   }
 `
```
alongoni, line 47: What about adding a empty `<HelperBox />` in line 80? So we can have this empty box/container and fill when we need it with some componetts (e.g, search, paginator, filter, etc) 
```
const HelperBox = styled.div`
  width: 100%;
  padding: 16px;
  gap: 16px;
  display: flex;
  align-items: center;
  justify-content: space-between;
`
```

henrypalacios: I will continue on this path because I've not been able to get the content of the tab to fill the entire space.

Selection_295

src/apps/explorer/components/common/ExplorerTabs/ExplorerTabs.stories.tsx
```diff
@@ -47,7 +48,35 @@ const tabItems = [
...
 
+const Wrapper = styled.div`
+  width: 500px;
+  padding: 16px;
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  border: 1px dashed darkorange;
+
+  .tab-extra-content {
```
henrypalacios, line 59: @alongoni Maybe you can help me [make this work](https://github.com/gnosis/gp-ui/blob/styling-tab-1043/src/components/common/Tabs/Tabs.tsx#L79).

src/apps/explorer/components/common/ExplorerTabs/ExplorerTab.tsx
```diff
@@ -33,6 +34,14 @@ const StyledTabs = styled.div`
   > div > div.tab-content {
     padding: 20px 16px;
   }
+
+  .tab-extra-content {
+    width: 100%;
+
+    @media ${MEDIA.mobile} {
+      display: none; /* for now we can hide the extra-content on mobiles */
```
alongoni, line 42: We should work on this in a new PR

src/components/common/Tabs/Tabs.tsx
```diff
@@ -58,10 +65,22 @@ export const DEFAULT_TAB_THEME: TabTheme = {
...
+}
+
+interface ExtraContentProps {
+  extra?: TabBarExtraContent
+}
+
+const ExtraContent = ({ extra }: ExtraContentProps): JSX.Element | null => {
+  if (!extra) return null
+
+  console.log('ExtraContent', extra)
```
alfetopito, line 78:
```suggestion
```

alfetopito (approved):
Looking good to me.

Any comments @biocom ?

biocom:
Looks good based on https://pr561--gpui.review.gnosisdev.com/storybook/?path=/story/explorerapp-explorertabs--default-tabs

One comment is, what happens for mobile/responsive. E.g. the search/filters and tabs would need a solution for that.

One idea is to turn the tabs into a select list/dropdown for mobile. Then column wrap all the items. Example https://ant.design/components/tabs/

Another idea is to only have these items:
- Sticky 'dropdown/select menu' on top (when scrolling).
- Sticky/fixed **FILTER* button at the bottom -> On click opens search/filtering options (similar to what you see on e-commerce websites). 

This would prioritize the initial order details on mobile, and only show the filters on request. I'd be in favor of the latter idea.

henrypalacios:
Due to last comment, what do you suggest @alfetopito? 
Should we keep it here or should I create another PR for this?

alfetopito:
> Due to last comment, what do you suggest @alfetopito?
> Should we keep it here or should I create another PR for this?

New PR. Since the desktop version is good to go, I'll merge it.

PR #569, opened 2021-08-06
Add TradesTable component

stackseeder:
Closes gnosis/cowswap#1045

Add a TradesTable component

To Test
https://pr569--gpui.review.gnosisdev.com/storybook/?path=/story/trades-tradestable--default
[ ] Under Trade should be a TradesTable section


src/components/trade/TradesTable/index.tsx
```diff
@@ -0,0 +1,176 @@
...
+import { OrderSurplusDisplay } from 'components/orders/OrderSurplusDisplay'
+import { calculatePrice, formatSmart, invertPrice, safeTokenName, TokenErc20 } from '@gnosis.pm/dex-js'
+import {
+  HIGH_PRECISION_DECIMALS,
+  HIGH_PRECISION_SMALL_LIMIT,
+  NO_ADJUSTMENT_NEEDED_PRECISION,
+} from 'apps/explorer/const'
+import BigNumber from 'bignumber.js'
+
+const Icon = styled(FontAwesomeIcon)`
```
alfetopito, line 21: I see you are re-using the same code from https://github.com/gnosis/gp-ui/blob/534de13ce79703d081e753e0052c97048c9a1815/src/components/orders/OrderPriceDisplay/index.tsx#L20

Great idea!

In that case, extract this as a component to re-use it on both :)

src/components/trade/TradesTable/index.tsx
```diff
@@ -0,0 +1,176 @@
...
+      <th>Type</th>
+      <th>Sell Amount</th>
+      <th>Buy Amount</th>
+      <th>
+        Limit price
+        <Icon icon={faExchangeAlt} onClick={(): void => tableContext.inverPrice()} />
+      </th>
+      <th>
+        Execution Price
+        <Icon icon={faExchangeAlt} onClick={(): void => tableContext.inverPrice()} />
```
alfetopito, line 46: You don't need to wrap `tableContext.inverPrice`, you can pass it directly

src/components/trade/TradesTable/index.tsx
```diff
@@ -0,0 +1,176 @@
...
+  color: ${({ theme }): string => theme.grey};
+  font-size: ${({ theme }): string => theme.fontSizeDefault};
+  font-weight: ${({ theme }): string => theme.fontNormal};
+  display: flex;
+  align-items: center;
+`
+
+export const TradesTable = ({ header, className, numColumns, data }: Props): JSX.Element => {
+  const tableContext = React.useContext(TradesTableContext)
+  const [rows, setRows] = React.useState(<></>)
```
alfetopito, line 72: I don't think it's a good approach to store a component on state.
If you must store something for performance reasons, store the calculated data, and build the component on the fly

src/components/trade/TradesTable/index.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+  React.useEffect((): void => {
+    // for each entry in the data array invert the values of
+    // the limit price and execution price if toggled
+    const _rows = data.map((order, i): JSX.Element => {
+      interface price {
+        amount: BigNumber
+        fraction: TokenErc20
```
alfetopito, line 80: Why name this `fraction`?

src/components/trade/TradesTable/index.tsx
```diff
@@ -0,0 +1,176 @@
...
+      }
+      interface priceData {
+        denominator: price
+        numerator: price
+      }
+
+      const constructPrice = (data: priceData): string => {
```
alfetopito, line 87: Same for this function, could probably be declared outside the component

src/components/trade/TradesTable/index.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+export const TradesTable = ({ header, className, numColumns, data }: Props): JSX.Element => {
+  const tableContext = React.useContext(TradesTableContext)
+  const [rows, setRows] = React.useState(<></>)
+
+  React.useEffect((): void => {
+    // for each entry in the data array invert the values of
+    // the limit price and execution price if toggled
+    const _rows = data.map((order, i): JSX.Element => {
+      interface price {
```
alfetopito, line 78: The interface is declared within a loop, not a good idea.

src/components/trade/TradesTable/index.tsx
```diff
@@ -0,0 +1,176 @@
...
+      }
+
+      return (
+        <tr key={i}>
+          <td>
+            <TitleUid textToCopy={order.uid} contentsToDisplay={order.shortId} />
+          </td>
+          <td className={i % 2 === 1 ? 'long' : 'short'}>{i % 2 === 1 ? 'Buy' : 'Sell'}</td>
+          <td>
+            {order.sellToken?.decimals.toFixed(3)} {order.sellToken?.symbol}
```
alfetopito, line 118: We want to display the amounts, not the decimals ;)
Also, we'll probably want to use `formatSmart` for the amounts. See how that's done on https://github.com/gnosis/gp-ui/blob/dfb80806a9577f9bbb223bc999fa9fd0e540e60b/src/components/orders/AmountsDisplay/index.tsx#L33

src/components/trade/TradesTable/index.tsx
```diff
@@ -0,0 +1,176 @@
...
+                  },
+                  numerator: {
+                    amount: order.buyAmount,
+                    fraction: order.buyToken,
+                  },
+                })
+              : '-'}
+          </td>
+          <td>
+            <TitleUid
```
alfetopito, line 152: Tx hash should link to etherscan, like https://github.com/gnosis/gp-ui/blob/36eae4570aac573e71d909042d3d4b18adfbf5aa/src/components/orders/DetailsTable/index.tsx#L169

src/components/trade/TradesTable/index.tsx
```diff
@@ -0,0 +1,176 @@
...
+            {order.buyToken && order.sellToken
+              ? constructPrice({
+                  denominator: {
+                    amount: order.buyAmount,
+                    fraction: order.buyToken,
+                  },
+                  numerator: {
+                    amount: order.sellAmount,
+                    fraction: order.sellToken,
+                  },
```
alfetopito, line 133: The order of denominator/numerator is inverted
screenshot_2021-08-06_08-48-48

alfetopito (changes requested):
Good work so far, left a couple of comments

alongoni (commented):
Looks really good! 
There are some minor css details. 
- Add a line between Table-head and Table-body
- Adjust the height of the rows to 50px

Heres is the design to compare:
https://www.figma.com/file/m7obOydxAhGNmlFYvxTk3Z/GP-Explorer?node-id=683%3A1280

src/components/orders/OrderPriceDisplay/index.tsx
```diff
@@ -47,28 +42,30 @@ export function OrderPriceDisplay(props: Props): JSX.Element {
...
+  const invert = (): void => setIsPriceInverted(!isPriceInverted)
 
-  const [baseSymbol, quoteSymbol] = isPriceInverted ? [sellSymbol, buySymbol] : [buySymbol, sellSymbol]
+  React.useEffect((): void => {
+    const price: string = constructPrice({
+      isPriceInverted: props.isPriceInverted ?? false,
+      order: props,
+      data: {
+        numerator: {
+          amount: new BigNumber(sellAmount),
```
alfetopito, line 54: sell/buyAmounts are already BigNumbers

src/components/orders/OrderPriceDisplay/index.tsx
```diff
@@ -47,28 +42,30 @@ export function OrderPriceDisplay(props: Props): JSX.Element {
...
-  })
-
-  const buySymbol = safeTokenName(buyToken)
-  const sellSymbol = safeTokenName(sellToken)
+  const [formattedPrice, setFormattedPrice] = useState('')
+  const invert = (): void => setIsPriceInverted(!isPriceInverted)
 
-  const [baseSymbol, quoteSymbol] = isPriceInverted ? [sellSymbol, buySymbol] : [buySymbol, sellSymbol]
+  React.useEffect((): void => {
+    const price: string = constructPrice({
```
alfetopito, line 49: I suggest to separate the price calculation from the price formatting for display purposes.

Also, you could probably use these helper functions to calculate the prices

https://github.com/gnosis/gp-ui/blob/36eae4570aac573e71d909042d3d4b18adfbf5aa/src/utils/operator.ts#L172 

https://github.com/gnosis/gp-ui/blob/36eae4570aac573e71d909042d3d4b18adfbf5aa/src/utils/operator.ts#L195

src/components/trade/TradesTable/index.stories.tsx
```diff
@@ -0,0 +1,96 @@
...
+      >
+        <Component {...rest} />
+      </TradesTableContext.Provider>
+    </div>
+  )
+}
+
+const order = {
+  ...RICH_ORDER,
+  buyAmount: new BigNumber('10000000000'), // 1WETH
```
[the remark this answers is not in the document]
alfetopito: Actually, you just help me find a bug on the test data.

RAW_ORDER has sell token WETH https://github.com/gnosis/gp-ui/blob/c89c25dc5d7fa612c8cd01c3d5046860a355f34f/test/data/operator.ts#L21

RICH_ORDER has sell token USDT https://github.com/gnosis/gp-ui/blob/c89c25dc5d7fa612c8cd01c3d5046860a355f34f/test/data/operator.ts#L52

Could you please fix that bug for me?

src/components/trade/TradesTable/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+                isPriceInverted: tableContext.isPriceInverted,
+                order,
+                data: {
+                  numerator: {
+                    amount: order.buyAmount,
+                    token: order.buyToken,
+                  },
+                  denominator: {
+                    amount: order.sellAmount,
+                    token: order.sellToken,
```
alfetopito, line 107: Numerator/denominator is inverted for execution price

src/components/trade/TradesTable/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+                  <tr key={i}>
+                    <td>
+                      <TitleUid textToCopy={order.uid} contentsToDisplay={order.shortId} />
+                    </td>
+                    <td className={i % 2 === 1 ? 'long' : 'short'}>{i % 2 === 1 ? 'Buy' : 'Sell'}</td>
+                    <td>
+                      {order.sellToken?.decimals.toFixed(3)} {order.sellToken?.symbol}
+                    </td>
+                    <td>
+                      {order.buyToken?.decimals.toFixed(3)} {order.buyToken?.symbol}
```
alfetopito, line 134: You are displaying the token decimals instead of the sell/buy amounts

src/components/trade/TradesTable/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+      body={
+        <>
+          {rows.length > 0 &&
+            rows.map(
+              (order: Order & { limitPrice: string; executionPrice: string }, i) =>
+                order.uid && (
+                  <tr key={i}>
+                    <td>
+                      <TitleUid textToCopy={order.uid} contentsToDisplay={order.shortId} />
+                    </td>
```
alfetopito, line 128: Can you link the order id to order details page?

src/components/trade/TradesTable/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+                    <td className={i % 2 === 1 ? 'long' : 'short'}>{i % 2 === 1 ? 'Buy' : 'Sell'}</td>
```
alfetopito, line 129: It's slightly different in the design.
I'm fine to address in another PR though. You can for now leave out the className thing.
And if you want to test inverting buy/sell, just change the data in the storybook :)

alfetopito (changes requested):

src/components/trade/TradesTable/index.stories.tsx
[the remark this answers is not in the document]
stackseeder: @alfetopito you mean sellToken and buyToken should be changed to each other?

src/components/trade/TradesTable/index.tsx
```diff
@@ -0,0 +1,249 @@
...
+  ${TableTheme}
+`
+
+export const TradesTableHeader = (): JSX.Element => {
+  const tableContext = React.useContext(TradesTableContext)
+
+  return (
+    <StyledTableRow variant={'default'}>
+      <th>
+        Tx Id <IconOutline icon={faQuestion} />
```
alongoni, line 73: We can reuse this icon?
https://protocol-explorer.dev.gnosisdev.com/rinkeby/orders/0x3bdceef7965b0f342aa37f7489007cd4ce2c9a07b485571351f133943d23d4cab6bad41ae76a11d10f7b0e664c5007b908bc77c961144d83
image

src/components/trade/TradesTable/index.tsx
```diff
@@ -0,0 +1,249 @@
...
+  const [tokens, setTokens] = React.useState<TokenErc20[]>([])
+  React.useEffect(() => {
+    // isLong - When True, determines if trade type is a BUY else trade type is a SELL
+    setTokens(isLong ? [buyToken, sellToken] : [sellToken, buyToken])
+  }, [isLong, buyToken, sellToken])
+
+  return tokens.length > 0 ? (
+    <TradeTypeWrapper>
+      <span>{isLong ? 'Buy' : 'Sell'}</span>&nbsp;
+      <TokenImg address={tokens[0].address} />
```
alongoni, line 117: Token icons should have 16px height and width.

alongoni (changes requested):

src/components/common/BlockExplorerLink/BlockExplorerLinkNetworkState.tsx
```diff
@@ -1,14 +1,15 @@
...
 
 import { BlockExplorerLink, Props } from './BlockExplorerLink'
 
 /**
  * BlockExplorerLink which relies on the network state
  */
 export function BlockExplorerLinkNetworkState(props: Props): JSX.Element {
-  const networkId = useNetworkId() || undefined
+  // const networkId = useNetworkId() || undefined
+  const networkId = useNetworkOrDefault() || undefined
```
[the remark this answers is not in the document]
stackseeder: The useNetworkId() function fails when it is unable to identify the current network but useNetworkOrDefault() has a fallback value. When used on the TradesTable to render the tx Hash value with a link, it doesn't work, except the useNetworkOrDefault() is used instead.

src/components/trade/TradesTable/index.stories.tsx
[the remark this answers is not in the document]
alfetopito: I mean RICH_ORDER sell/buyToken is inverted. Just switch them

src/components/common/BlockExplorerLink/BlockExplorerLinkNetworkState.tsx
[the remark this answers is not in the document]
alfetopito: Yes, it makes sense to use a fallback value.
I just wonder why that works for other pages where this is used.
Maybe that's an issue with the storybook?

Try adding the `NetworkDecorator` https://github.com/gnosis/gp-ui/blob/1bec4a89d70d90cc6ff59d8fd7a9c5656167130a/src/components/orders/DetailsTable/DetailsTable.stories.tsx#L18

test/data/operator.ts
```diff
@@ -48,8 +48,8 @@ export const RICH_ORDER: Order = {
   fullyFilled: false,
   filledAmount: ZERO_BIG_NUMBER,
   filledPercentage: ZERO_BIG_NUMBER,
-  buyToken: WETH,
-  sellToken: USDT,
+  buyToken: USDT,
+  sellToken: WETH,
```
alfetopito, line 52: Thanks!

src/components/trade/TradesTable/index.tsx
```diff
@@ -0,0 +1,234 @@
...
+                      ) : (
+                        '-'
+                      )}
+                    </td>
+                    <td>{!order.surplusAmount.isZero() ? <OrderSurplusDisplay order={order} /> : '-'}</td>
+                    <td>
+                      {numberWithCommas(order.sellAmount)} {order.sellToken?.symbol}
+                    </td>
+                    <td>
+                      {numberWithCommas(order.buyAmount)} {order.buyToken?.symbol}
```
alfetopito, line 217: Use formatSmartMaxPrecision

anxolin (commented):
Thanks for the PR @stackseeder, i see some progress there, however im a bit concerned about two points:
- The PR was submitted for review either too early judging by all the issues in the code and the big difference to the table. 
- The PR also feels a bit big

To be considerate with the reviewers time, we need to make it easy for them. This means, to:
- **Hold the PR until it meets the quality standard**. Otherwise, people need to read the PR very very careful and is not clear what is finished and what is not
- **If the PR is big, we don't need to do all at once**. You can break down big tasks in smaller ones, and clearly explain in the PR the scope of your change. This makes it easier to the reviewer to understand what he/she need to verify. 

I think if you do this, we will give you a better support, and would help our dev process. 

Thanks! 🙏🙏

src/components/trade/TradesTable/index.tsx
alongoni: The token icons are too small (10px). We need to grow them to 16px.
image

src/api/operator/types.ts
```diff
@@ -17,7 +17,7 @@ export interface FeeInformation {
   feeRatio: number
 }
 
-export type OrderKind = 'sell' | 'buy'
+export type OrderTradeKind = 'sell' | 'buy'
```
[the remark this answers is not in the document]
stackseeder: I've fixed the issue.

src/api/operator/types.ts
```diff
@@ -82,26 +82,33 @@ export type RawTrade = {
   logIndex: number
   owner: string
   txHash: string
-  orderUid: string
+  tradeId: string
```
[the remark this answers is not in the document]
stackseeder: I'v fixed the issue

src/api/operator/types.ts
```diff
@@ -82,26 +82,33 @@ export type RawTrade = {
...
+  kind: OrderTradeKind
 }
 
 /**
  * Enriched Trade type
  */
-export type Trade = Pick<RawTrade, 'blockNumber' | 'logIndex' | 'owner' | 'txHash'> & {
-  orderId: string // rename the field
+export type Trade = Pick<RawTrade, 'blockNumber' | 'logIndex' | 'owner' | 'txHash' | 'kind'> & {
+  tradeId: string
```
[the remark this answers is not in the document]
stackseeder: I've fixed the issue.

elena-zh:
Order id is displayed as 'TX ID'
image

Also, in the main task #1045 there is a requirement to show TxHash, but it is missing in the current table (or Order ID is missing, I'm not sure)

alongoni:
I think we can improve the copy button message to prevent this behavior:
https://www.loom.com/share/b42080b7591c493190d58d8573a272f4

henrypalacios:
@alongoni that's solved [here](https://github.com/gnosis/gp-ui/blob/c0ceee12951e71d78ea18644dd67905cbe948edd/src/components/common/StyledUserDetailsTable.tsx#L49), just add the correct class in [RowWithCopyButton](https://github.com/gnosis/gp-ui/pull/569/files#diff-2f42b919466c8083b85f2702a8521540359fe08ef7bcb05a1aadfa00c11dd596R85) @ivica.

src/api/operator/types.ts
```diff
@@ -84,24 +84,37 @@ export type RawTrade = {
...
   sellAmount: string
+  executedSellAmount: string
+  executedFeeAmount: string
   sellAmountBeforeFees: string
   buyToken: string
   sellToken: string
+  executionTime: string
+  kind: OrderKind
+  surplusAmount: string
+  surplusPercentage: string
```
alfetopito, line 97: Where is this data coming from?
They are not part of the api response.
Please make sure this type contains only data from the API.

alfetopito (changes requested):
Please see Elena's comment above regarding the columns.
Take a look at the columns in the issue for reference
screenshot_2021-08-31_13-36-21

When you are done and wants a new review, re-request the review (click on screenshot_2021-08-31_13-39-18) from involved parties to indicate it's ready

screenshot_2021-08-31_13-38-09

src/api/operator/types.ts
stackseeder: @alfetopito I've confused the fields from the orders api response.
I will remove those fields. But then I have some questions.
- Since the trades API does not return the fields like executedBuyAmount, executedSellAmount, executedFeeAmount, how we can calculate the executionPrice value?
- Also surplus does not exist in the api response. Then where should we get the value from?
- The same question for executionTime.

alfetopito: Yes, that's something I know it's a problem.
For now, the plan is to get those values from individual `/order/<id>` calls for every single trade.
Not optimal, but that's the current db design.

For this reason, @henrypalacios 's idea to have a context shared between orders and trades table is super useful.
You can look up the given orderId on the context first and only if it doesn't exist query the API endpoint.

Anyway, I'll have a meeting with backend guys tomorrow to see if we can improve the data returned on these queries, but for the trades endpoint I don't expect much will change.

stackseeder: @alfetopito then what do you think if we integrate the trades table with the actual API in this PR and close the other PR?
I guess it would be unnecessary to mock the order/trade data and fetch orders from the mock.
I could be not helpful to simplify the PR rather than integrating with real API.

alfetopito: I still think they should be separated.
This PR is meant to have the UI only, using storybook, independent of the data source.

You can use static data as input on the stories. There's no need for a mock as far as I can tell. 

You define what's the interface the UI component will have, and  - even if the data from the API is not in the given format - you can transform it before passing it to the component.

Keep the UI data representation isolated from the API data representation. So both can be updated independently.

Also, this PR is not too far from being done.
I think once you have all the fields required it should be good to merge.

alongoni (approved):
Looks great!

src/components/trade/TradesTable/index.tsx
```diff
@@ -0,0 +1,186 @@
...
+  const { executionTime, buyToken, buyAmount, sellToken, sellAmount, kind, orderId } = trade
+
+  return (
+    <tr key={orderId}>
+      <td>
+        {
+          <RowWithCopyButton
+            className="span-copybtn-wrap"
+            textToCopy={orderId}
+            contentsToDisplay={<Link to={`/trades/${trade.orderId}`}>{orderId}</Link>}
```
alfetopito, line 109: Point the link to the order details page, and display the shortened order id https://github.com/gnosis/gp-ui/blob/434997a0260c7e8fb663ba675c70bcba01166a81/src/utils/operator.ts#L228

```suggestion
            contentsToDisplay={<Link to={`/orders/${trade.orderId}`}>{getShortOrderId(orderId)}</Link>}
```

src/components/trade/TradesTable/index.tsx
```diff
@@ -0,0 +1,186 @@
...
+  max-width: 10rem;
+  overflow: hidden;
+  text-overflow: ellipsis;
+`
+
+function isTokenErc20(token: TokenErc20 | null | undefined): token is TokenErc20 {
+  return (token as TokenErc20).address !== undefined
+}
+
+function formattedAmount(erc20: TokenErc20 | null | undefined, amount: BigNumber): string {
```
alfetopito, line 46: These helper fns are pretty the same as https://github.com/gnosis/gp-ui/blob/1dfe1e7702d505c64d62460b7f1574d40d47124e/src/components/orders/OrdersUserDetailsTable/index.tsx#L29 Henry created, right?

Could you move them to a common utils file?

alfetopito (commented):
Looking really good!
Some minor comments

src/components/trade/TradesTable/index.tsx
```diff
@@ -0,0 +1,176 @@
...
+  })
+
+  return formatCalculatedPriceToDisplay(calculatedPrice, trade.buyToken, trade.sellToken, isPriceInverted)
+}
+
+function getExecutedPrice(trade: Trade, isPriceInverted: boolean): string {
+  if (!trade.buyToken || !trade.sellToken) return '-'
+
+  const order: RawOrder = {
+    ...RAW_ORDER,
```
alfetopito, line 60: I missed this in the last review.
Please don't use test data inside regular components.
Use them only on unit tests or stories

I understand you need a RawOrder instance as a parameters to `getOrderExecutedPrice` and you don't have that here.
Then I suggest to update the function params to pick only the required fields https://github.com/gnosis/gp-ui/blob/434997a0260c7e8fb663ba675c70bcba01166a81/src/utils/operator.ts#L146

Such as:

```ts
export type GetRawOrderPriceParams = CommonPriceParams & {
  order: Pick<RawOrder, 'executedSellAmount' | '...' >
}
```

Then you can keep the same interface but do not need to build a RawOrder instance.

src/components/trade/TradesTable/index.tsx
```diff
@@ -0,0 +1,176 @@
...
+  isPriceInverted: boolean
+}
+
+const RowOrder: React.FC<RowProps> = ({ trade, isPriceInverted }) => {
+  const { executionTime, buyToken, buyAmount, sellToken, sellAmount, kind, orderId } = trade
+
+  return (
+    <tr key={orderId}>
+      <td>
+        {
```
alfetopito, line 95: Do you need to wrap the component in `{}` here?

src/components/trade/TradesTable/index.tsx
```diff
@@ -1,21 +1,21 @@
...
+import { BlockExplorerLink } from 'components/common/BlockExplorerLink'
 import {
-  formatSmartMaxPrecision,
   getOrderLimitPrice,
   getOrderExecutedPrice,
   formatCalculatedPriceToDisplay,
   formatExecutedPriceToDisplay,
+  formattedAmount,
 } from 'utils'
+import { getShortOrderId } from '../../../utils/operator'
```
alfetopito, line 18: Use the full path

elena-zh:
Hey @alfetopito , @stackseeder , can we move TxHash to the end of the table ? 
Or, at least, closer to Order ID? 
In this case, I think, it's will be easier to find, and trade details will look a bit more consistent without splitting numbers with link. 

Also, it would be good to add 'Copy' icon to the TxHash like it works for an order details page
image

I'm not sure whether there was a decision to implement a responsive view of this table or not under the current task, but I'd better mention that the table looks pretty strange without fields' labels and with scrolling header
image

I have to mention, that the same behavior is for Orders table.

Thanks

elena-zh:
Hey @stackseeder , great!
The only thing is that 'Copy' icon for TxHash column copies an order ID instead. 
Could you please fix it?

stackseeder:
> Hey @stackseeder , great!
> The only thing is that 'Copy' icon for TxHash column copies an order ID instead.
> Could you please fix it?

@elena-zh please check that again. I've fixed it.

elena-zh:
Now changes LGTM. Thanks!

alfetopito:
> Hey @alfetopito , @stackseeder , can we move TxHash to the end of the table ?
> Or, at least, closer to Order ID?
> In this case, I think, it's will be easier to find, and trade details will look a bit more consistent without splitting numbers with link.

Good idea Elena

> Also, it would be good to add 'Copy' icon to the TxHash like it works for an order details page
> image

No strong opinion here, ok to have it

> I'm not sure whether there was a decision to implement a responsive view of this table or not under the current task, but I'd better mention that the table looks pretty strange without fields' labels and with scrolling header
> image
> 
> I have to mention, that the same behavior is for Orders table.

Indeed. I proposed for the first round to not worry about responsiveness, to make the scope and the PRs smaller.
As I mentioned during our sync today, it'll come as a follow up.

alfetopito (approved):
Excellent, I'm quite happy with the result :+1:

anxolin (approved):
Looks good in desktop. This PR has been open too long. I think can be merged, and reiterated.

Other than the responsive, one nitpick, is that I find strange how our copy functionality moves the content. I wonder if the "Copied" should just float above so it doesn't create that effect.

The table content, looks great. Looking forward to have a release that include this table.

alongoni:
> Other than the responsive, one nitpick, is that I find strange how our copy functionality moves the content. I wonder if the "Copied" should just float above so it doesn't create that effect.

I've the same feeling with the copy button. Maybe we can extend the [copy button component](https://pr569--gpui.review.gnosisdev.com/storybook/?path=/story/common-copybutton--default) with a variant showing the copy message in a tooltip.

elena-zh:
Btw, there is an issue for this component: https://github.com/gnosis/gp-ui/issues/630
May be, we can change its description with a new decision how it can look like.

PR #583, opened 2021-08-11
1044 orders userdetails table

henrypalacios:
closes: gnosis/cowswap#1044

:spiral_notepad: **TODO**:
- [ ] Integrate 2 commons from [1044_1045](https://github.com/gnosis/gp-ui/pull/582) proxy branch
- [ ] Add `Button` invertPrice

### DIFF
--- a/src/components/orders/OrdersUserDetailsTable/index.stories.tsx
+++ b/src/components/orders/OrdersUserDetailsTable/index.stories.tsx
@@ -1,0 +1,43 @@
+import React from 'react'
+// also exported from '@storybook/react' if you can deal with breaking changes in 6.1
+import { Story, Meta } from '@storybook/react/types-6-0'
+import OrderUserDetailsTable, { Props as OrdersUserTableProps } from '.'
+import { add, sub } from 'date-fns'
+import BigNumber from 'bignumber.js'
+import { GlobalStyles, ThemeToggler, NetworkDecorator } from 'storybook/decorators'
+
+import { Order } from 'api/operator'
+import { RICH_ORDER, TUSD, WETH } from '../../../../test/data'
+
+export default {
+  title: 'orders/OrdersUserDetailsTable',
+  decorators: [GlobalStyles, NetworkDecorator, ThemeToggler],
+  component: OrderUserDetailsTable,
+} as Meta
+
+const orderBuy: Order = {
+  ...RICH_ORDER,
+  kind: 'buy',
+  buyToken: TUSD,
+  sellToken: WETH,
+  shortId: 'bdef89ac',
+  buyAmount: new BigNumber('1000000000000000000'), // 1WETH
+  sellAmount: new BigNumber('3000000000'), //3000 USDT
+  creationDate: sub(new Date(), { hours: 1 }),
+  expirationDate: add(new Date(), { hours: 1 }),
+  txHash: '0x489d8fd1efd43394c7c2b26216f36f1ab49b8d67623047e0fcb60efa2a2c420b',
+}
+
+const orderSell: Order = {
+  ...RICH_ORDER,
+  buyAmount: new BigNumber('1000000000000000000'), // 1WETH
+  sellAmount: new BigNumber('3000000000'), //3000 USDT
+  creationDate: sub(new Date(), { hours: 1 }),
+  expirationDate: add(new Date(), { hours: 1 }),
+  txHash: '0x489d8fd1efd43394c7c2b26216f36f1ab49b8d67623047e0fcb60efa2a2c420b',
+}
+
+const Template: Story<OrdersUserTableProps> = (args) => <OrderUserDetailsTable {...args} />
+
+export const Default = Template.bind({})
+Default.args = { orders: [orderBuy, orderSell], showBorderTable: true }
--- a/src/components/orders/OrdersUserDetailsTable/index.tsx
+++ b/src/components/orders/OrdersUserDetailsTable/index.tsx
@@ -1,0 +1,104 @@
+import React from 'react'
+import styled from 'styled-components'
+
+import { Order } from 'api/operator'
+
+import StyledUserDetailsTable, { StyledUserDetailsTableProps } from './styled'
+import { DateDisplay } from 'components/orders/DateDisplay'
+import { RowWithCopyButton } from 'components/orders/RowWithCopyButton'
+import { TokenDisplay } from 'components/common/TokenDisplay'
+import { formatSmartMaxPrecision } from 'utils'
+import { TokenErc20 } from '@gnosis.pm/dex-js'
+import BigNumber from 'bignumber.js'
+import { StatusLabel } from '../StatusLabel'
+import { Network } from 'types'
+
+const Wrapper = styled(StyledUserDetailsTable)`
+  > thead > tr,
+  > tbody > tr {
+    grid-template-columns: 12rem 17rem repeat(4, 13rem) 1fr;
+  }
+`
+function isTokenErc20(token: TokenErc20 | null | undefined): token is TokenErc20 {
+  return (token as TokenErc20).address !== undefined
+}
+
+const formattedAmount = (erc20: TokenErc20 | null | undefined, amount: BigNumber): string => {
+  if (!isTokenErc20(erc20)) return '-'
+
+  return erc20.decimals ? formatSmartMaxPrecision(amount, erc20) : amount.toString(10)
+}
+
+const getLimitPrice = (order: Order): string => {
+  if (!order.buyToken || !order.sellToken) return '-'
+
+  return `${formattedAmount(order.sellToken, order.sellAmount)} ${order.sellToken?.symbol} per ${
+    order.sellToken?.symbol
+  }`
+}
+
+export type Props = StyledUserDetailsTableProps & {
+  orders: Order[]
+}
+
+const OrdersUserDetailsTable: React.FC<Props> = (props) => {
+  const { orders, showBorderTable = false } = props
+  const network = Network.Mainnet
+
+  const orderItems = (items: Order[]): JSX.Element => {
+    return (
+      <>
+        {items.map((item) => (
+          <tr key={item.shortId}>
+            <td>
+              {
+                <RowWithCopyButton
+                  className="span-copybtn-wrap"
+                  textToCopy={item.uid}
+                  contentsToDisplay={item.shortId}
+                />
+              }
+            </td>
+            <td className="cell-kind-operation">
+              {item.kind} {<TokenDisplay erc20={item.sellToken as TokenErc20} network={network} />}
+              for {<TokenDisplay erc20={item.buyToken as TokenErc20} network={network} />}
+            </td>
+            <td>
+              {formattedAmount(item.sellToken, item.sellAmount)} {item.sellToken?.symbol}
+            </td>
+            <td>
+              {formattedAmount(item.buyToken, item.buyAmount)} {item.buyToken?.symbol}
+            </td>
+            <td>{getLimitPrice(item)}</td>
+            <td>
+              <StatusLabel status={item.status} partiallyFilled={item.partiallyFilled} />
+            </td>
+            <td>
+              <DateDisplay date={item.creationDate} />
+            </td>
+          </tr>
+        ))}
+      </>
+    )
+  }
+
+  return (
+    <Wrapper
+      showBorderTable={showBorderTable}
+      header={
+        <tr>
+          <th>Order ID</th>
+          <th>Type</th>
+          <th>Buy amount</th>
+          <th>Sell amount</th>
+          <th>Limit price</th>
+          <th>Status</th>
+          <th>Creation time</th>
+        </tr>
+      }
+      body={orderItems(orders)}
+    />
+  )
+}
+
+export default OrdersUserDetailsTable
--- a/src/components/orders/OrdersUserDetailsTable/styled.tsx
+++ b/src/components/orders/OrdersUserDetailsTable/styled.tsx
@@ -1,0 +1,49 @@
+import styled from 'styled-components'
+
+import { SimpleTable, Props as SimpleTableProps } from 'components/common/SimpleTable'
+
+interface Props {
+  showBorderTable?: boolean
+}
+
+export type StyledUserDetailsTableProps = SimpleTableProps & Props
+
+const StyledUserDetailsTable = styled(SimpleTable)<StyledUserDetailsTableProps>`
+  border: ${({ theme, showBorderTable }): string => (showBorderTable ? `0.1rem solid ${theme.borderPrimary}` : 'none')};
+  border-radius: 0.4rem;
+
+  tr td {
+    &:not(:first-of-type) {
+      text-align: right;
+    }
+
+    &.long {
+      border-left: 0.2rem solid var(--color-long);
+    }
+
+    &.short {
+      color: var(--color-short);
+      border-left: 0.2rem solid var(--color-short);
+    }
+  }
+
+  thead tr th {
+    color: white;
+    font-style: normal;
+    font-weight: 800;
+    font-size: 13px;
+    line-height: 16px;
+    height: 50px;
+    border-bottom: ${({ theme }): string => `1px solid ${theme.borderPrimary}`};
+  }
+
+  tbody tr:hover {
+    backdrop-filter: contrast(0.9);
+  }
+
+  .span-copybtn-wrap {
+    display: block;
+  }
+`
+
+export default StyledUserDetailsTable


henrypalacios:
I would like to clarify some doubts @alfetopito:

1. Should I create a link to `/orders/<uid>` in **Order ID** [column](https://github.com/gnosis/gp-ui/blob/1044-orders-userdetails-table/src/components/orders/OrdersUserDetailsTable/index.tsx#L77)?
2. Should I use a formatting like that of the component `AmountDetails` [formattedAmount function](https://github.com/gnosis/gp-ui/blob/1044-orders-userdetails-table/src/components/orders/AmountsDisplay/index.tsx#L33)?
3. Should I display always the `buyToken`/`buyAmount` in **Buy amount** column? 
Or is it okay to exchange it when it's an sell order (`kind == 'sell'`)
Selection_300

src/components/orders/OrdersUserDetailsTable/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+  return (token as TokenErc20).address !== undefined
+}
+
+const formattedAmount = (erc20: TokenErc20 | null | undefined, amount: BigNumber): string => {
+  if (!isTokenErc20(erc20)) return '-'
+
+  return erc20.decimals ? formatSmartMaxPrecision(amount, erc20) : amount.toString(10)
+}
+
+const getLimitPrice = (order: Order): string => {
```
henrypalacios, line 37: @alfetopito I have used as reference the logic in `OrderPriceDisplay` [component](https://github.com/gnosis/gp-ui/blob/1044-orders-userdetails-table/src/components/orders/OrderPriceDisplay/index.tsx#L52), could it be taken out to reuse here?

alfetopito:
> I would like to clarify some doubts @alfetopito:
> 
>     1. Should I create a link to `/orders/<uid>` in **Order ID** [column](https://github.com/gnosis/gp-ui/blob/1044-orders-userdetails-table/src/components/orders/OrdersUserDetailsTable/index.tsx#L77)?

Yes!
 
>     2. Should I use a formatting like that of the component `AmountDetails` [formattedAmount function](https://github.com/gnosis/gp-ui/blob/1044-orders-userdetails-table/src/components/orders/AmountsDisplay/index.tsx#L33)?

Yep
 
>     3. Should I display always the `buyToken`/`buyAmount` in **Buy amount** column?

You mean the symbol with the amount? E.g.: `0.123123 ETH`
I agree it's redundant with the other info in the table, but we have in the past settled to be explicit.

Unless @biocom has a different opinion, I say we keep it.

>        Or is it okay to exchange it when it's an sell order (`kind == 'sell'`)
>        Selection_300

I don't quite get the second part. Buy and sell amounts should always be visible, if that's what you are asking.

src/components/orders/OrdersUserDetailsTable/index.tsx
alfetopito: Yes, that's a good idea.

Actually, @stackseeder is working on exactly the same thing, you guys are overlapping.

I commented on his PR https://github.com/gnosis/gp-ui/pull/569#discussion_r687239936

So my suggestion is either:
1. Use the function as you have it, extracting and updating OrdersPriceDisplay
2. Use the utils functions implemented and create another utils for formatting the price

I wrote those helpers (https://github.com/gnosis/gp-ui/blob/36eae4570aac573e71d909042d3d4b18adfbf5aa/src/utils/operator.ts#L172 and https://github.com/gnosis/gp-ui/blob/36eae4570aac573e71d909042d3d4b18adfbf5aa/src/utils/operator.ts#L195) but I cannot remember now why I did not use them.

alfetopito: Also, just noticed that your code does not account for the price being inverted, which should be possible to do for the whole table at once.

henrypalacios:
> I don't quite get the second part. Buy and sell amounts should always be visible, if that's what you are asking.

Let me give you more context @alfetopito, when viewing the complete table:
ordersTable

- When the order is to buy, the **Buy amount** column makes sense with what the token I am Buying.
- But when is a sell order, I am not sure what goes in the **Buy Amount** column. Logic tells me that am buying is the other pair :sweat_smile: , (i.e. DAI, according to the image).

alfetopito:
> > I don't quite get the second part. Buy and sell amounts should always be visible, if that's what you are asking.
> 
> Let me give you more context @alfetopito, when viewing the complete table:
> ordersTable
> 
>     * When the order is to buy, the **Buy amount** column makes sense with what the token I am Buying.
> 
>     * But when is a sell order, I am not sure what goes in the **Buy Amount** column. Logic tells me that am buying is the other pair sweat_smile , (i.e. DAI, according to the image).
> 
> 
> My question is, if it is ok as it is in the previous image, or if it should be DAI?

Now I get it.

Let me explain what's the difference:

The most common orders are of type SELL.
When you go to CowSwap and input an amount in the FROM field:
screenshot_2021-08-13_14-48-07
That is saying: I want to sell EXACTLY this amount. How much do I get? (buyAmount).
For these orders, the buyAmount is variable, and can even get more than the initial amount. That's the surplus for SELL orders.

On BUY orders, they are created when you input an amount in the TO field:
screenshot_2021-08-13_14-49-02
That is saying: I want to buy EXACTLY this amount. How much do I have to sell to get that? (sellAmount)
For these orders, the sellAmount is variable, and can be less than your initial price. The surplus in this case is in the form of paying less for the same amount.

Is it more clear?

---

Back to the question of what to show, you'll always show buy and sell amounts as you get them in the order.
Every order will always have one amount being sold and another being bought.

Regarding the image you shared, the values in the last row are swapped. BUY amount is in DAI and SELL amount is in WETH.

elena-zh:
Hey @henrypalacios , @alfetopito !
I'd like to add some enhancements here. Could we add for
sell order: 'at least'
buy order: 'at most'
image

Also, according to the discussion in https://github.com/gnosis/cowswap/issues/1135#issuecomment-899451694 , it would be nice to apply 4-decimals rounding for amounts in user's order table

alfetopito:
> Hey @henrypalacios , @alfetopito !
> I'd like to add some enhancements here. Could we add for
> sell order: 'at least'
> buy order: 'at most'
> image


> Also, according to the discussion in [gnosis/cowswap#1135 (comment)](https://github.com/gnosis/cowswap/pull/1135#issuecomment-899451694) , it would be nice to apply 4-decimals rounding for amounts in user's order table

For this, it would be better another issue so we address the precision throughout the app at once.

biocom:
Couple of comments:
- I'd omit adding the 'at least' and 'at most' in the same cell as the digits. This will mess with any (future) sorting logic of the column. Also I'd argue it makes it harder to visually 'scan' the table overview. An alternative to still accommodate for this, could be to add a toggle to show min. received amount instead.
- I would consider splitting the `buy` and `sell` in the TYPE column. Reason for this is that it will be easier to filter out BUY/SELL orders and allows you to sort on them if desired. I would even go as far as removing the `TUSD for WETH` given that's redundant. 
- I think it's good practice to move the `STATUS` column as the last one. Easier to scan order status and some will have actions tied to them (cancel order).

- Example:

| Order Id | Type | Sell amount       | Buy amount | Limit price                 | Created                          | Status |
|----------|------|-------------------|------------|-----------------------------|----------------------------------|--------|
| bd304832 |  Buy | 0.0000000003 WETH |   1 TUSD   | < 0.000000001 WETH per TUSD | 1 hour ago 08/16/2021 3:51:25 PM | Open   |
| adef89ad | Sell | 1 WETH            | 3,000 USDT | 3,000 USDT per WETH         | 1 hour ago 08/16/2021 3:51:25 PM | Open   |


Other example where `SELL AMOUNT` and `BUY AMOUNT` are consolidated:

| Order Id | Type | Amount                               | Limit price                 | Created                          | Status |
|----------|------|--------------------------------------|-----------------------------|----------------------------------|--------|
| bd304832 |  Buy | 1 TUSD for at most 0.0000000003 WETH | < 0.000000001 WETH per TUSD | 1 hour ago 08/16/2021 3:51:25 PM | Open   |
| adef89ad | Sell | 1 WETH for at least 3,000 USDT       | 3,000 USDT per WETH         | 1 hour ago 08/16/2021 3:51:25 PM | Open   |


Personally I think the last example suffices for the purpose of Order Management UX. Our (inline) search and filters could narrow down orders in case you want to filter by a specific token and/or amount.

henrypalacios:
@alfetopito According to last comment, I would like to define how the table would look like? 
Like the first example?
Selection_310

or like the second one?
Selection_309

If you choose any of these, how would the icon look like?
Selection_308

src/utils/operator.ts
```diff
@@ -153,13 +153,17 @@ export function getOrderExecutedAmounts(order: RawOrder): {
...
 
-export type GetOrderPriceParams = {
+export type GetRawOrderPriceParams = {
   order: RawOrder
   buyTokenDecimals: number
   sellTokenDecimals: number
   inverted?: boolean
 }
 
+export interface GetOrderPriceParams extends Omit<GetRawOrderPriceParams, 'order'> {
```
henrypalacios, line 163: I had to add an extension to the property in `getOrderLimitPrice` to be able to us in [`OrdersUsersDetailsTable`](https://github.com/gnosis/gp-ui/blob/1044-orders-userdetails-table/src/components/orders/OrdersUserDetailsTable/index.tsx#L45). 

@alfetopito maybe we could use here only what is used inside which is `amount`, `token` and not the whole order.

alfetopito:
> @alfetopito According to last comment, I would like to define how the table would look like?
> Like the first example?
> Selection_310
> 
> or like the second one?
> Selection_309
> 
> If you choose any of these, how would the icon look like?
> Selection_308

I personally prefer the first option.
And I vote for not having token symbols at all.

Unless @biocom has a different view on this?

@anxolin , wanna be the tie breaker?

src/utils/operator.ts
alfetopito: That suggestion makes a lot of sense.
Feel free to update the function to expect only those parameters instead of the whole order object.

As far as I remember it won't break anything as the function is not currently in use.

Also, now I remember why I wrote that function. 
I intended to use it when building the order object and have the limit price available (and execution price if there was a trade) as part of the Order object.
I ended up not using it because based only on the Order data is not possible to calculate the price, since we need both token decimals which are fetched on another call.

biocom:
I agree with @alfetopito 's points :
- 1st option of the table is good
- No token symbol on the BUY or SELL. Open to use the token symbol next to the token name (e.g. what we do on GP tx detail, but instead only use the token symbol and not full name)
Screen Shot 2021-08-19 at 08 25 37

alfetopito:
@henrypalacios as mentioned during our sync call, when I said no symbols I meant no "icons", sorry about that.

So following up on Michel's comment, `type` should only be `sell` or `buy`.
Sell/buy `amounts` should have the token `symbols` as a suffix.

Pretty much the same as this table:

> Selection_310

anxolin:
Catching up in this thread. Agree with all the last changes suggested to the table. Looks way better now. 

Just nitpicking, since mostly is done for using it imo:

1. Ethercan, has some icon next to the date. Maybe we want to consider adding it (fell free to ignore this one if it doesn't look good).
image

2. Normally **BUY** is represented in green, and **SELL** in red.
image

3. It would be nice to add one example for each state. For example, I would be curious about `FILLED` state. Are we sure we won't add any of the filds like: `Surplus` or `Execution price`. Should the `Limit Price` be just called `Price` so it's Limit/Execution depending if it's dilled or not?

4. All orders are `fill or kill` for now, but will we make in this view any difference with the partially fillables? 

5. I guess the responsive part of this one is handle separately. 

Anyways, very nice progress! I'm looking forward to see this one in action! 💪

alfetopito:
> Catching up in this thread. Agree with all the last changes suggested to the table. Looks way better now.
> 
> Just nitpicking, since mostly is done for using it imo:
> 
>     1. Ethercan, has some icon next to the date. Maybe we want to consider adding it (fell free to ignore this one if it doesn't look good).
>        image
> 
>     2. Normally **BUY** is represented in green, and **SELL** in red.
>        image

Not 100% convinced that would work here, but not against either.

>     3. It would be nice to add one example for each state. For example, I would be curious about `FILLED` state. Are we sure we won't add any of the filds like: `Surplus` or `Execution price`. Should the `Limit Price` be just called `Price` so it's Limit/Execution depending if it's dilled or not?
> 
>     4. All orders are `fill or kill` for now, but will we make in this view any difference with the partially fillables?

I think this can be addressed later as limit orders are not yet a high priority. I prefer to focus on getting this feature out only for `fill-or-kill` first.

> 
>     5. I guess the responsive part of this one is handle separately.

Yep

> 
> 
> Anyways, very nice progress! I'm looking forward to see this one in action! muscle

henrypalacios:
@anxolin Thank you for your suggestions.

**1. and 2.**
---

Applied, but may need a finer eye evaluation @biocom. Or small quick design @alongoni 
Selection_314

**3. and 4.**
---
I'm not sure if something else should change when there are other states in the table @alfetopito.
Otherwise maybe I can force the dummy data or I could create a history for the [StatusLabel](https://github.com/gnosis/gp-ui/blob/1044-orders-userdetails-table/src/components/orders/StatusLabel/index.tsx#L62) component  where the different states would be appreciated.

There are things that escape me here at the moment, I would need more context to give an assessment.

alfetopito:
Oh, I missed the point `3`

>> It would be nice to add one example for each state. For example, I would be curious about FILLED state. Are we sure we won't add any of the filds like: Surplus or Execution price. Should the Limit Price be just called Price so it's Limit/Execution depending if it's dilled or not?

> I'm not sure if something else should change when there are other states in the table @alfetopito.
Otherwise maybe I can force the dummy data or I could create a storybook for the StatusLabel component where the different states would be appreciated.

Not now, but in the short term we intend to allow orders to be cancelled, so the pending state `Open` would have a cancel button. (See one of the examples from Michel)
Expired and cancelled orders have no addition.
Filled state, hmm, could have `surplus` next to the state.

As for the other things, I initially though about them only for the OrdersTrades table.
For additional details the user can view them by opening the order details page.

I would say, we can start with the table as is and iterate with what we might want in follow up PRs based on feedback.

What do you say @anxolin ?

alongoni:
> @anxolin Thank you for your suggestions.
> 
> ## **1. and 2.**
> Applied, but may need a finer eye evaluation @biocom. Or small quick design @alongoni
> Selection_314
> 
Regarding 1. and 2. Looks good! I prefer the clock icon in light mode (left option). Maybe we can omit the () parenthesis?

src/utils/format.ts
```diff
@@ -242,3 +247,36 @@ export function formatSmartMaxPrecision(amount: BigNumber, token?: TokenErc20 |
...
+    precision: NO_ADJUSTMENT_NEEDED_PRECISION,
+    smallLimit: HIGH_PRECISION_SMALL_LIMIT,
+    decimals: HIGH_PRECISION_DECIMALS,
+  })
+  const buySymbol = safeTokenName(buyToken)
+  const sellSymbol = safeTokenName(sellToken)
+
+  const [baseSymbol, quoteSymbol] = isPriceInverted ? [sellSymbol, buySymbol] : [buySymbol, sellSymbol]
+
+  return `${formattedPrice} ${quoteSymbol} per ${baseSymbol}`
```
alfetopito, line 281: Just realized this is inverted.

Not sure though whether it's here or in the input data for the story.

screenshot_2021-08-20_08-30-47

If you are selling `3000 USDT` and buing `1 WETH`, you'll have the price to be `3000 WETH/USDT` (standard notation, can be shortened to just `3000 WETH`) that is the same as `USDT per WETH` (extended notation).

src/components/orders/OrdersUserDetailsTable/index.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { RICH_ORDER, TUSD, WETH } from '../../../../test/data'
+
+export default {
+  title: 'orders/OrdersUserDetailsTable',
+  decorators: [Router, GlobalStyles, NetworkDecorator, ThemeToggler],
+  component: OrderUserDetailsTable,
+} as Meta
+
+const orderBuy: Order = {
+  ...RICH_ORDER,
```
alfetopito, line 19: It's possible the inverted price issue I mentioned in the other comment to be related to this bug https://github.com/gnosis/gp-ui/blob/c89c25dc5d7fa612c8cd01c3d5046860a355f34f/test/data/operator.ts#L21-L52
Ivica fixed it on his PR. The sell/buy tokens are inverted in the `RICH_ORDER`

src/utils/format.ts
alfetopito: If you invert the price, you should have the `0.003... WETH per USDT`

src/components/orders/OrdersUserDetailsTable/index.stories.tsx
henrypalacios: It was a really a trouble maker :smile:

PR #611, opened 2021-08-26
1046 load from api orderstable

henrypalacios:
Closes gnosis/cowswap#1046

Load orders from a wallet address.

:spiral_notepad: **Summary**:
- [x] Get the API to fetch data.
- [x] Refresh orders every 30s on the background.
- [x] Filter by time ago to save on calls. 
- [x] Orders Loader

## To Test
1. Go to  https://pr611--gpui.review.gnosisdev.com/rinkeby/address/0xb6BAd41ae76A11D10f7b0E664C5007b908bC77C9/
2. No orders. https://pr611--gpui.review.gnosisdev.com/address/0xb6BAd41ae76A11D10f7b0E664C5007b908bC77C9/

### DIFF
--- a/src/api/operator/operatorApi.ts
+++ b/src/api/operator/operatorApi.ts
@@ -193,16 +193,24 @@ export async function getOrder(params: GetOrderParams): Promise<RawOrder | null>
  *  - owner: address
  *  - sellToken: address
  *  - buyToken: address
+ *  - minValidTo: number
  */
 export async function getOrders(params: GetOrdersParams): Promise<RawOrder[]> {
   const { networkId, ...searchParams } = params
-  const { owner, sellToken, buyToken } = searchParams
+  const { owner, sellToken, buyToken, minValidTo } = searchParams
+  const defaultValues = {
+    includeFullyExecuted: 'true',
+    includeInvalidated: 'true',
+    includeInsufficientBalance: 'true',
+    includePresignaturePending: 'true',
+    includeUnsupportedTokens: 'true',
+  }
 
   console.log(
     `[getOrders] Fetching orders on network ${networkId} with filters: owner=${owner} sellToken=${sellToken} buyToken=${buyToken}`,
   )
 
-  const searchString = buildSearchString({ ...searchParams })
+  const searchString = buildSearchString({ ...searchParams, ...defaultValues, minValidTo: String(minValidTo) })
 
   const queryString = '/orders/' + searchString
 
--- a/src/apps/explorer/components/OrdersTableWidget/useGetOrders.tsx
+++ b/src/apps/explorer/components/OrdersTableWidget/useGetOrders.tsx
@@ -20,7 +20,7 @@ export function mergeNewOrders(previousOrders: Order[], newOrdersFetched: RawOrd
 
   // find the order up to which it is to be replaced
   const lastOrder = newOrdersFetched[newOrdersFetched.length - 1]
-  const positionLastOrder = previousOrders.map((o) => o.uid).indexOf(lastOrder.uid)
+  const positionLastOrder = previousOrders.findIndex((o) => o.uid === lastOrder.uid)
   if (positionLastOrder === -1) {
     return newOrdersFetched.map((order) => transformOrder(order)).concat(previousOrders)
   }
@@ -73,7 +73,7 @@ export function useGetOrders(ownerAddress: string): Result {
         }, [])
 
         setErc20Addresses(newErc20Addresses)
-        // For the moment it is neccesary to sort by date
+        // TODO -> For the moment it is neccesary to sort by date
         ordersFetched.sort((a, b) => +new Date(b.creationDate) - +new Date(a.creationDate))
 
         setOrders((previousOrders) => mergeNewOrders(previousOrders, ordersFetched))
--- a/src/apps/explorer/components/OrdersTableWidget/useGetOrders.tsx
+++ b/src/apps/explorer/components/OrdersTableWidget/useGetOrders.tsx
@@ -1,0 +1,128 @@
+import { useState, useCallback, useEffect } from 'react'
+import { subMinutes, getTime } from 'date-fns'
+
+import { Network } from 'types'
+import { useMultipleErc20 } from 'hooks/useErc20'
+import { getOrders, Order, RawOrder } from 'api/operator'
+import { useNetworkId } from 'state/network'
+import { transformOrder } from 'utils'
+import { ORDERS_HISTORY_MINUTES_AGO, ORDERS_QUERY_INTERVAL } from 'apps/explorer/const'
+
+/**
+ *
+ * Merge new RawOrders consulted, that may have changed status
+ *
+ * @param previousOrders List of orders
+ * @param newOrdersFetched List of fetched block order that could have changed
+ */
+export function mergeNewOrders(previousOrders: Order[], newOrdersFetched: RawOrder[]): Order[] {
+  if (newOrdersFetched.length === 0) return previousOrders
+
+  // find the order up to which it is to be replaced
+  const lastOrder = newOrdersFetched[newOrdersFetched.length - 1]
+  const positionLastOrder = previousOrders.map((o) => o.uid).indexOf(lastOrder.uid)
+  if (positionLastOrder === -1) {
+    return newOrdersFetched.map((order) => transformOrder(order)).concat(previousOrders)
+  }
+
+  const slicedOrders: Order[] = previousOrders.slice(positionLastOrder + 1)
+  return newOrdersFetched.map((order) => transformOrder(order)).concat(slicedOrders)
+}
+
+function isObjectEmpty(object: Record<string, unknown>): boolean {
+  // eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars
+  for (const key in object) {
+    if (key) return false
+  }
+
+  return true
+}
+
+type Result = {
+  orders: Order[]
+  error: string
+  isLoading: boolean
+}
+
+export function useGetOrders(ownerAddress: string): Result {
+  const [isLoading, setIsLoading] = useState(false)
+  const [error, setError] = useState('')
+  const [orders, setOrders] = useState<Order[]>([])
+  const networkId = useNetworkId() || undefined
+  const [erc20Addresses, setErc20Addresses] = useState<string[]>([])
+  const { value: valueErc20s, isLoading: areErc20Loading } = useMultipleErc20({ networkId, addresses: erc20Addresses })
+  const [mountNewOrders, setMountNewOrders] = useState(false)
+
+  const fetchOrders = useCallback(
+    async (network: Network, owner: string, minTimeHistoryTimeStamp = 0): Promise<void> => {
+      setIsLoading(true)
+      setError('')
+
+      try {
+        const ordersFetched = await getOrders({ networkId: network, owner, minValidTo: minTimeHistoryTimeStamp })
+        const newErc20Addresses = ordersFetched.reduce((accumulator: string[], element) => {
+          const updateAccumulator = (tokenAddress: string): void => {
+            if (accumulator.indexOf(tokenAddress) === -1) {
+              accumulator.push(tokenAddress)
+            }
+          }
+          updateAccumulator(element.buyToken)
+          updateAccumulator(element.sellToken)
+
+          return accumulator
+        }, [])
+
+        setErc20Addresses(newErc20Addresses)
+        // For the moment it is neccesary to sort by date
+        ordersFetched.sort((a, b) => +new Date(b.creationDate) - +new Date(a.creationDate))
+
+        setOrders((previousOrders) => mergeNewOrders(previousOrders, ordersFetched))
+        setMountNewOrders(true)
+      } catch (e) {
+        const msg = `Failed to fetch orders`
+        console.error(msg, e)
+        setError(msg)
+      } finally {
+        setIsLoading(false)
+      }
+    },
+    [],
+  )
+
+  useEffect(() => {
+    if (!networkId) {
+      return
+    }
+    const getOrUpdateOrders = (minHistoryTime?: number): Promise<void> =>
+      fetchOrders(networkId, ownerAddress, minHistoryTime)
+
+    getOrUpdateOrders()
+
+    const intervalId: NodeJS.Timeout = setInterval(() => {
+      const minutesAgoTimestamp = getTime(subMinutes(new Date(), ORDERS_HISTORY_MINUTES_AGO))
+      getOrUpdateOrders(Math.floor(minutesAgoTimestamp / 1000))
+    }, ORDERS_QUERY_INTERVAL)
+
+    return (): void => {
+      clearInterval(intervalId)
+    }
+  }, [fetchOrders, networkId, ownerAddress])
+
+  useEffect(() => {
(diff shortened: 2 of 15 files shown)


henrypalacios:
@alfetopito The problem I had with the API is that I need to extend the back date. 
Selection_323

Also when when I was testing in Swagger-Rinkeby I was using mainnet.
Selection_324

### I would like to know, when you mentioned:
>Probably only includeFullyExecuted=false to exclude filled orders and includeInvalidated=false to exclude cancelled orders.
1. I'm not sure what query strings to will be used to display orders in [`OrdersTable`](https://github.com/gnosis/gp-ui/blob/1046-load-from-api-orderstable/src/apps/explorer/components/OrdersTableWidget/OrdersTableWithData.tsx#L105) by default, because as I imagined it was to show the past orders [including the executed](https://pr611--gpui.review.gnosisdev.com/rinkeby/address/0xb6BAd41ae76A11D10f7b0E664C5007b908bC77C9/).  
 (I used default values for show data like a [oneYearAgo](https://github.com/gnosis/gp-ui/blob/1046-load-from-api-orderstable/src/api/operator/operatorApi.ts#L200)).
 


2. I had to go looking for the token [contracts for the orders](https://github.com/gnosis/gp-ui/blob/1046-load-from-api-orderstable/src/apps/explorer/components/OrdersTableWidget/OrdersTableWithData.tsx#L39), How about using a context here in [`OrdersTableWidget`](https://github.com/gnosis/gp-ui/blob/1046-load-from-api-orderstable/src/apps/explorer/components/OrdersTableWidget/index.tsx#L41) to hold the multiple ERC20 contracts used in `OrdersTable` as `TradesTable`?

alfetopito:
> @alfetopito The problem I had with the API is that I need to extend the back date.
> Selection_323

Well, I was sure I used this before, but it seems I mistook /orders with /trades endpoint.
Yeah, `minValidTo` must be set in the past to show older orders.
By default to start with all trades we could set this to `0`.
Subsequent calls we perform the same query with `minValidTo` set to 10 min ago or something like that, then update the existing orders.

Also, set all boolean fields to `true` such as https://protocol-rinkeby.dev.gnosisdev.com/api/v1/orders?owner=0x5B0ABE214aB7875562ADeE331dEfF0Fe1912FE42&includeFullyExecuted=true&includeInvalidated=true&includeInsufficientBalance=true&includePresignaturePending=true&includeUnsupportedTokens=true&minValidTo=0

We want to show everything on the explorer.

> 
> Also when when I was testing in Swagger-Rinkeby this was using was mainnet.
> Selection_324

You can switch on top which endpoint you are calling. You can access all of them from the same page. I know, weird.
screenshot_2021-08-31_14-13-07


> ### I would like to know, when you mentioned:
> 
> > Probably only includeFullyExecuted=false to exclude filled orders and includeInvalidated=false to exclude cancelled orders.
> 
>1. I'm not sure what query strings to will be used to display orders in [`OrdersTable`](https://github.com/gnosis/gp-ui/blob/1046-load-from-api-orderstable/src/apps/explorer/components/OrdersTableWidget/OrdersTableWithData.tsx#L105) by default, because as I imagined it was to show the past orders [including the executed](https://pr611--gpui.review.gnosisdev.com/rinkeby/address/0xb6BAd41ae76A11D10f7b0E664C5007b908bC77C9/).

I've answered this on the top above. Set all flags to true and start with `minValidTo=0`

>        (I used default values for show data like a [oneYearAgo](https://github.com/gnosis/gp-ui/blob/1046-load-from-api-orderstable/src/api/operator/operatorApi.ts#L200)).
> 
>2. I had to go looking for the token [contracts for the orders](https://github.com/gnosis/gp-ui/blob/1046-load-from-api-orderstable/src/apps/explorer/components/OrdersTableWidget/OrdersTableWithData.tsx#L39), How about using a context here in [`OrdersTableWidget`](https://github.com/gnosis/gp-ui/blob/1046-load-from-api-orderstable/src/apps/explorer/components/OrdersTableWidget/index.tsx#L41) to hold the multiple ERC20 contracts used in `OrdersTable` as `TradesTable`?

Yes, that makes a lot of sense. It'll be very handy as well for @stackseeder since trades have even less context and need the respective orders in order to build the full row of data.

The tokens are already stored on global state (once loaded). Orders and trades are not because it only makes sense to do so if viewing the user details page so your idea of a context is great for the user details page.

src/api/operator/operatorApi.ts
```diff
@@ -197,12 +197,19 @@ export async function getOrder(params: GetOrderParams): Promise<RawOrder | null>
 export async function getOrders(params: GetOrdersParams): Promise<RawOrder[]> {
   const { networkId, ...searchParams } = params
   const { owner, sellToken, buyToken } = searchParams
+  const oneYearAgo = (): string => {
+    const currentDate = new Date()
+    const yearAgo = new Date(currentDate.setMonth(currentDate.getMonth() - 1))
+
+    return String(Math.floor(yearAgo.getTime() / 1000))
+  }
```
alfetopito, line 205: We probably won't need this function, but anyway, it doesn't seem to do what I think it should:
screenshot_2021-08-31_14-38-14

Also, we have https://date-fns.org/ as a dependency, makes our lives easier :)

elena-zh:
Hey @henrypalacios , @alfetopito ,
I might be a bit lost, but is seems to me that we do not have 'Search', 'Filtering' and 'Pagination' components for a user's orders/trades table
search

elena-zh:
I know, that this PR is WIP, but I would like to add a suggestion to add a default sort order to tables (created desc or so)
image

and to display see/buy amounts better or shorter..
image

alfetopito:
> Hey @henrypalacios , @alfetopito ,
> I might be a bit lost, but is seems to me that we do not have 'Search', 'Filtering' and 'Pagination' components for a user's orders/trades table
> search

We agreed to not have search, filtering and pagination on the first version.

> I have an endless table for my Rinkeby account with lines overlapping the table header, that is why I raised this issue
> https://pr611--gpui.review.gnosisdev.com/rinkeby/address/0xFF714b8b0e2700303eC912BD40496C3997ceEa2b/
> 
> image

Hmm, I do not see any orders when I open this link :/

alfetopito:
> I know, that this PR is WIP, but I would like to add a suggestion
> 
>1. to add a default sort order to tables (created desc or so)
>        image

Orders are supposed to come directly from the API sorted by creation date descending.
For now, they are not. I'll have a sync with backend team tomorrow to discuss this.

>2. to display see/buy amounts better or shorter..
>        image

Agree this should be improved. But I think we can do it all at once, similar to what's been done on CowSwap.
It'll involve updating how Order details are displayed as well so better to do in a separated task.

>3. to app a loader when order details are loaded
>        https://drive.google.com/file/d/1gYDU8PymA690htXpTuf0SQg6Wz4BZLXU/view

:+1: 

>4. to open an order link in a new tab

:+1: 

>     5. Can we move the `\/ Copied` to a new row?
>        image

Not a big deal to me right now. Maybe we could replace the text with 'copied' depending on the screen size.
But again, also another task.

>     6. Will we add a sorting feature to columns?

Yes, but not on the first version.

elena-zh:
> Hmm, I do not see any orders when I open this link :/

I was not able to open this page yesterday, today I see orders again. So here is the video: https://drive.google.com/file/d/1VpDeAdbZHNziW-xyqKNVEGiTwgL37jP2/view

Btw, looks a bit weird when order list flashing when a user's order info is updated.

elena-zh:
I have created separate issues for cases: 
 - move the `\/ Copied` to a new row #630 
 - enhance amounts displaying #629

henrypalacios:
@elena-zh 
> I was not able to open this page yesterday  

you're right, I had problems with this, I thought it was my changes, but then I realized that something had changed in the API.

> Btw, looks a bit weird when order list flashing when a user's order info is updated.

I'm working on it 😅 , I just wanted to make sure the table is updated. 

And thank you for your other remarks.

src/apps/explorer/const.ts
```diff
@@ -2,6 +2,8 @@ import { AnalyticsDimension } from 'types'
 
 /** Explorer app constants */
 export const ORDER_QUERY_INTERVAL = 10000 // in ms
+export const ORDERS_QUERY_INTERVAL = 10000 // in ms
```
henrypalacios, line 5: We had agreed on 30min but I have set it to 10min to test.

alongoni:
Looks great!
We need to add the Title: "User details"
Screenshot from 2021-09-09 11-09-06
"

src/apps/explorer/components/OrdersTableWidget/useGetOrders.tsx
```diff
@@ -0,0 +1,128 @@
...
+ *
+ * @param previousOrders List of orders
+ * @param newOrdersFetched List of fetched block order that could have changed
+ */
+export function mergeNewOrders(previousOrders: Order[], newOrdersFetched: RawOrder[]): Order[] {
+  if (newOrdersFetched.length === 0) return previousOrders
+
+  // find the order up to which it is to be replaced
+  const lastOrder = newOrdersFetched[newOrdersFetched.length - 1]
+  const positionLastOrder = previousOrders.map((o) => o.uid).indexOf(lastOrder.uid)
```
alfetopito, line 23: map+indexOf you are going over the list of orders twice.
Can do a single pass with findIndex (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex)

src/apps/explorer/components/OrdersTableWidget/useGetOrders.tsx
```diff
@@ -0,0 +1,128 @@
...
+            }
+          }
+          updateAccumulator(element.buyToken)
+          updateAccumulator(element.sellToken)
+
+          return accumulator
+        }, [])
+
+        setErc20Addresses(newErc20Addresses)
+        // For the moment it is neccesary to sort by date
```
alfetopito, line 76: Add a TODO here to not forget to remove later

src/apps/explorer/components/OrdersTableWidget/useGetOrders.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+  // find the order up to which it is to be replaced
+  const lastOrder = newOrdersFetched[newOrdersFetched.length - 1]
+  const positionLastOrder = previousOrders.map((o) => o.uid).indexOf(lastOrder.uid)
+  if (positionLastOrder === -1) {
+    return newOrdersFetched.map((order) => transformOrder(order)).concat(previousOrders)
+  }
+
+  const slicedOrders: Order[] = previousOrders.slice(positionLastOrder + 1)
+  return newOrdersFetched.map((order) => transformOrder(order)).concat(slicedOrders)
```
alfetopito, line 29: Although this is ok for now, I think the assumption regarding the data ordering will be wrong when the query is paginated.

I mean, right now the query returns all results.
With pagination, in page X we'll get only existing orders, so concating them this way will create duplicates/overwrite newer orders.

Anyway, to keep in mind

alfetopito (commented):
Working quite nicely!

Left a few refactoring suggestions.

And, this isn't part of the logic but I missed when reviewing the table design.
On full screen the columns don't fill the whole table
screenshot_2021-09-09_14-26-38

src/apps/explorer/components/OrdersTableWidget/useGetOrders.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+  const fetchOrders = useCallback(
+    async (network: Network, owner: string, minTimeHistoryTimeStamp = 0): Promise<void> => {
+      setIsLoading(true)
+      setError('')
+
+      try {
+        const ordersFetched = await getOrders({ networkId: network, owner, minValidTo: minTimeHistoryTimeStamp })
+        const newErc20Addresses = ordersFetched.reduce((accumulator: string[], element) => {
+          const updateAccumulator = (tokenAddress: string): void => {
```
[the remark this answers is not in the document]
henrypalacios: I love the Set, it would be much more readable however I would go for the reduce for performance. 
Do you think it makes up for it? https://jsben.ch/ognIg

elena-zh:
Hey @henrypalacios , great pr!
A new tiny issue from me: ann empty orders table is jumping when updating each 30 secs 
https://drive.google.com/file/d/14c41B_FzYnQHCu4zEk6XoI9Z-tFeqB3M/view

Also, btw, can we change 'Canceled' status to 'Cancelled' to match Cowswap app?
I know that it is a known pending issue under the #612 task, but may be it is an easy fix?

elena-zh:
Hey @henrypalacios , changes you implemented look great!
But still orders table is jumping each 30 sec when it is empty:
https://drive.google.com/file/d/1bTWypM3bHfc4OHVbfUrrraXgxiL5JvQe/view
Also, it looks a bit weird when hovered state is applied to an empty table
image

Thanks!

src/apps/explorer/components/OrdersTableWidget/useGetOrders.tsx
henrypalacios: Ok, this won't be necessary, I have removed it in the [PR-661](https://github.com/gnosis/gp-ui/pull/661) where I work with the new endpoint `/account/{owner}/orders`

src/apps/explorer/components/OrdersTableWidget/useGetOrders.tsx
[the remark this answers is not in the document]
alfetopito: Wow, no idea it had such a performance impact. Guess they make mose sense in languages where Sets are first class citizens.

I prefer sets for readability but maybe we should stick with reducer due to performance as you suggest.

@anxolin @W3stside FYI

test/utils/operator/orderFormatAmount.test.ts
```diff
@@ -0,0 +1,42 @@
+import BigNumber from 'bignumber.js'
+
+import { isTokenErc20, formattedAmount } from 'utils'
+
+import { TokenErc20 } from '@gnosis.pm/dex-js'
+
+const factoryToken = (): TokenErc20 => {
```
alfetopito, line 7: Minor nitpick: any reason for not being a constant instead of a function?

alfetopito (approved):
Love it! Working great!

test/utils/operator/orderFormatAmount.test.ts
henrypalacios: I was going to create a **TokenErc20** factory, for preference I wanted a concise syntax, however it is not necessary to make an object creator because it will not require variation.

henrypalacios:
@elena-zh has been removed in the row when there are not orders.

@alongoni has suggested using the same size of the table when **reloading** so it doesn't look strange.

@alfetopito I have set the interval back to 30sec. I also think something is going wrong in the evaluation of the test coverage.

alongoni (approved):
Looks great!

alfetopito:
I force merged it.
The failure is due to an outage on Coveralls https://status.coveralls.io/ unrelated to your changes

elena-zh:
Hey @henrypalacios , great!
 
Just curious: why do I always see these 404 errors in the console? 
image

In addition, please, don't kill me.. But still I see blinking table when it is updating each 30 seconds (see from 19 sec )
https://drive.google.com/file/d/1DFNOWTSpBvpr8hJgzFAuiIOH9OvkFxO0/view

A bit strange to see this scroll in the empty table. It is also displayed when a table has orders, but it is more logical to show it there, but not in the empty table.. 
image

PR #615, opened 2021-08-27
Hotfix 2.2.4

alfetopito:
# Summary

HOTFIX!

Only 1 tiny change:

- Unblocks display of partially fillable orders #605 

# To Test 

Needs a partially fillable order and the right environemnt.
There's one placed on PROD so it can only be tested on Explorer PROD or STAGING.


W3stside (commented):
apprewv

PR #631, opened 2021-09-03
revert to BlockExplorerLink passed in 554

henrypalacios:
# Summary

I have made a change in [`DetailsTable`](https://github.com/gnosis/gp-ui/blob/hotfix-avoid-page-not-ready/src/components/orders/DetailsTable/index.tsx#L89) that was not convenient because it can go to production without the `UserDetailsPage` being finished.

### DIFF
--- a/src/components/orders/DetailsTable/index.tsx
+++ b/src/components/orders/DetailsTable/index.tsx
@@ -1,7 +1,6 @@
 import React from 'react'
 import styled from 'styled-components'
 import { media } from 'theme/styles/media'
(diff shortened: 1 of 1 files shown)


elena-zh (approved):
Great!

alfetopito:
As mentioned over slack, this change is not necessary at the moment.

The next planned release includes the user page currently linked, and should come out only when it's finished.

I see no issues leaving it pointing to the black page on dev for now.

We'll reconsider if we need to do a release without that.

anxolin (approved):

PR #657, opened 2021-09-14
[Sentry] source-maps support

nenadV91:
# Summary
This will provide source maps support in sentry console which will make easier to debug errors in production.

131670516-1d3644ea-8ce5-4a55-9dfd-7b4e127ad6c5

# To Test
- source maps support should be visible in the sentry console after the deployment of source code with this valid sentry authToken

# Background
- the sentry-webpack-plugin will only be added if the REACT_APP_SENTRY_AUTH_TOKEN is defined as a env variable. The value of this variable is the authToken provided by sentry.

### DIFF
--- a/src/apps/explorer/ExplorerApp.tsx
+++ b/src/apps/explorer/ExplorerApp.tsx
@@ -27,7 +27,7 @@ const SENTRY_TRACES_SAMPLE_RATE = process.env.REACT_APP_SENTRY_TRACES_SAMPLE_RAT
 
 if (SENTRY_DSN) {
   Sentry.init({
-    dsn: process.env.REACT_APP_SENTRY_DSN,
+    dsn: SENTRY_DSN,
     integrations: [new Integrations.BrowserTracing()],
     release: 'gp-explorer@v' + version,
     environment: environmentName,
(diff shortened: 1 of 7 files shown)


alfetopito (approved):

src/apps/explorer/ExplorerApp.tsx
```diff
@@ -25,11 +25,14 @@ import { version } from '../../../package.json'
 const SENTRY_DSN = process.env.REACT_APP_SENTRY_DSN
 const SENTRY_TRACES_SAMPLE_RATE = process.env.REACT_APP_SENTRY_TRACES_SAMPLE_RATE
 
+console.log('React app sentry dsn', SENTRY_DSN)
+console.log('GOOGLE_ANALYTICS_ID', process.env.GOOGLE_ANALYTICS_ID)
```
anxolin, line 29: Do we want to log these? Happy to do so if we really want. just asking

anxolin (approved):

src/apps/explorer/ExplorerApp.tsx
nenadV91: No this was just for debugging why there are no sentry issues for gp-ui. This will be removed.

Issue #684, opened 2021-09-22
Order table: empty table is blinking each 30 sec when reloading


elena-zh:
Originally reported in https://github.com/gnosis/gp-ui/issues/611#issuecomment-923974007
Related to https://github.com/gnosis/gp-ui/issues/661

I see blinking table when it is updating each 30 seconds (see from 19 sec )
https://drive.google.com/file/d/1DFNOWTSpBvpr8hJgzFAuiIOH9OvkFxO0/view

henrypalacios:
@elena-zh what you are suggesting is that you only show it the first time when the page loads?
Or remove it completely?

elena-zh:
I'd not to show it at all on UI. And start to reload it when the 1st order appears, IMO

Issue #693, opened 2021-09-27
Responsive Orders table


alfetopito:
**Description**

Make the Orders table responsive, as per design

**Proposal**

![screenshot_2021-09-27_14-15-39](https://user-images.githubusercontent.com/43217/134986379-8e5e8f7b-c65c-4a23-b13f-54be50f7669b.png)
(I don't know how to disable the red lines :facepalm:)

https://www.figma.com/file/m7obOydxAhGNmlFYvxTk3Z/GP-Explorer?node-id=280%3A1

elena-zh:
I noticed that there is no date on an order details in a mobile view. 
Are we OK with this?
I have to notice, that in the Cowswap sidebar we always show 'valid to' date (open/cancelled/expired) and 'filled on' (filled) for orders.
![image](https://user-images.githubusercontent.com/70885163/135069182-e29a100d-5c2a-44b7-8b47-06a2f84c8f62.png)

alfetopito:
Good point Elena.
As mentioned during the call, @alongoni will work on that.
I'd start with only creation date. Execution date is not yet available from the API AFAICT.

In the mean time if someone starts working on it just re-use the same component for the date like what's is used in other places on the Explorer:

`Creation date           xxx xxx xxx`

alongoni:
I've updated the design (the "creation date / created" column was added during the implementation 😅)
![UI_Explorer v1 - User details page-_ MOBILE](https://user-images.githubusercontent.com/622217/135158123-46b310d9-91a4-4216-b140-76c1df61a373.png)
)

elena-zh:
@alongoni , thank you!
As far as I understand, we will show a tooltip with the full date adjusted to the local time zone when click on the date. Correct?

alongoni:
> @alongoni , thank you! As far as I understand, we will show a tooltip with the full date adjusted to the local time zone when click on the date. Correct?

Right! I'm working improving this things and I'll sync with @matextrem in order to have all ready to adapt the design,

PR #694, opened 2021-09-29
Create buttons component to paginate user orders table

### LINKED ISSUE #687: Create a component to use as pager in user orders table
**Description**
Create buttons component to use as pager in the order table.
Selection_296

**Proposal**
- Create a button component with arrows.
- Use a component in top tab header.
- Pass limit filter to table when selecting a value.

**Additional context**
- [Figma proposal](https://www.figma.com/file/m7obOydxAhGNmlFYvxTk3Z/GP-Explorer?node-id=683%3A1280)

henrypalacios:
# Summary

Closes #687, #684 

- [x] Create a button component with arrows.
- [x] Use a component in top tab header with a legend of page count.
- [x] Handle button click to paginate the next or previous page.

## To Test
1. (687-Pager) -> Go to  https://pr694--gpui.review.gnosisdev.com/rinkeby/address/0xFF714b8b0e2700303eC912BD40496C3997ceEa2b/

2. (684-EmptyTable) Go to  https://pr694--gpui.review.gnosisdev.com/address/0xb6BAd41ae76A11D10f7b0E664C5007b908bC77C9/

### Reference

- [Figma proposal](https://www.figma.com/file/m7obOydxAhGNmlFYvxTk3Z/GP-Explorer?node-id=683%3A1280)

### DIFF
--- a/src/apps/explorer/components/OrdersTableWidget/PaginationOrdersTable.tsx
+++ b/src/apps/explorer/components/OrdersTableWidget/PaginationOrdersTable.tsx
@@ -37,6 +37,9 @@ const PaginationDropdownButton = styled.button`
     opacity: 0.5;
     pointer-events: none;
   }
+  &:hover span {
+    color: ${({ theme }): string => theme.textActive1};
+  }
 `
 
 const PaginationText = styled.p`
@@ -81,32 +84,63 @@ const PaginationButton = styled.button`
       color: ${({ theme }): string => theme.textActive1};
     }
   }
+
   &[disabled],
   &[disabled]:hover {
     cursor: not-allowed;
     opacity: 0.5;
-
     .fill {
       color: ${({ theme }): string => theme.textSecondary1};
     }
   }
 `
+PaginationButton.defaultProps = { disabled: true }
 
 const PaginationOrdersTable: React.FC = () => {
   const {
-    isFirstLoading,
-    tableState: { pageSize, pageOffset },
+    isOrdersLoading: isLoading,
+    tableState: { pageSize, pageOffset, canNextPage },
     setPageSize,
-    orders,
+    setPageOffset,
+    orders: rows,
   } = useContext(OrdersTableContext)
   const quantityPerPage = [10, 20, 30, 50]
+
+  const handleNextPage = (): void => {
+    const newOffset = pageOffset + pageSize
+    setPageOffset(newOffset)
+  }
+  const handlePreviousPage = (): void => {
+    const newOffset = pageOffset - pageSize
+    setPageOffset(newOffset < 0 ? 0 : newOffset)
+  }
+  const renderPageLegend = (): string => {
+    if (isLoading && !rows.length) return '.. - ..'
+
+    let startPageCount = 0
+    let endPageCount = 0
+    if (rows.length) {
+      startPageCount = pageOffset + 1
+      endPageCount = pageOffset + rows.length
+    }
+
+    return `${startPageCount} - ${endPageCount}`
+  }
+  const canPreviousPage = (): boolean => {
+    if (isLoading) return false
+    return pageOffset > 0 ? true : false
+  }
 
   return (
     <PaginationWrapper>
       <PaginationText>Rows per page:</PaginationText>
       <DropdownPagination
-        disabled={isFirstLoading}
-        dropdownButtonContent={<PaginationDropdownButton>{pageSize} ▼</PaginationDropdownButton>}
+        disabled={isLoading}
+        dropdownButtonContent={
+          <PaginationDropdownButton>
+            {pageSize} <span>▼</span>
+          </PaginationDropdownButton>
+        }
         dropdownButtonContentOpened={
           <PaginationDropdownButton className="selected">{pageSize} ▲</PaginationDropdownButton>
         }
@@ -117,13 +151,11 @@ const PaginationOrdersTable: React.FC = () => {
           </PaginationItem>
         ))}
       />
-      <PaginationText className="legend">
-        {pageOffset + 1 === 1 ? 1 : pageOffset * pageSize + 1} - {orders.length && pageOffset + orders.length}
-      </PaginationText>{' '}
-      <PaginationButton disabled={true} onClick={(): void => console.log('Left')}>
+      <PaginationText className="legend">{renderPageLegend()}</PaginationText>{' '}
+      <PaginationButton disabled={!canPreviousPage()} onClick={handlePreviousPage}>
         <Icon icon={faChevronLeft} className="fill" />
       </PaginationButton>
-      <PaginationButton disabled={!true} onClick={(): void => console.log('Right')}>
+      <PaginationButton disabled={!canNextPage} onClick={handleNextPage}>
         <Icon icon={faChevronRight} className="fill" />
       </PaginationButton>
     </PaginationWrapper>
--- a/src/apps/explorer/components/OrdersTableWidget/PaginationOrdersTable.tsx
+++ b/src/apps/explorer/components/OrdersTableWidget/PaginationOrdersTable.tsx
@@ -1,5 +1,7 @@
 import React, { useContext } from 'react'
 import styled, { css } from 'styled-components'
+// import { faChevronRight, faChevronLeft } from '@fortawesome/free-solid-svg-icons'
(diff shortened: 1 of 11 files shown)


elena-zh:
Hi @henrypalacios , changes LGTM!
I have found several minor issues, but I reported them in separate tasks that might be addressed later: 
#698  - enhancement to add navigation to the 1st/last pages, add Qty of lines
#699 - some blinking issues with pagination
#700 - issue with number of lines

Btw, pagination is missing in the mobile view, but I hope, that it will be addressed under #693

alongoni (approved):
Looks good Henry!

I agree with Elena's comment to keep it in the radar for future improvements.

alfetopito (commented):
Some minor comments on the code. Overall quite good :+1: 

Regarding the behaviour, echoing comments I made in the other issues I feel like make part of this change:

1. Changing the rows per page should reset position back to page 1
2. When not on page `1`, rows should not be added/removed, only update what's already visible. Now it causes rows to "shift" down with background updates. Probably the easiest is to disable background re-fresh when not on first page.
3. Not sold on showing the count of orders (51 - 100). Or maybe I'm missing the page number. What about adding that in between the page controls

For follow ups/other issues

4. Continuing the second page refresh issues. Keep track of what was the last order is the previous table and continue from there, avoiding showing the same order again.
5. Also on that, we could add a indicator like "new orders, click here to refresh" sort of thing. Low priority, though.
6. Add page controls to bottom of table as well - not part of the design but in practice I see the need

---

All in all, with `1` and `2` added it'll be good to merge IMO. We can discuss the rest and address in follow up tasks/PRs if we agree.

What do you think?

henrypalacios:
Thanks for the feedback @alfetopito .

I agree, I wanted to make progress on 1 and 2. But I got stuck on 2. Because I didn't realize that I was testing on `Rinkeby` at `https://cowswap.exchange/#/swap` 🤕.

alfetopito:
> Thanks for the feedback @alfetopito .
> 
> I agree, I wanted to make progress on 1 and 2. But I got stuck on 2. Because I didn't realize that I was testing on `Rinkeby` at `https://cowswap.exchange/#/swap` face_with_head_bandage. I'll work on point 2.

Just to be clear, Rinkeby works on any stage.
The issue in your case is that you were creating backend PROD orders.
Using anything other than PROD will give you STAGING backend orders.
So, barn.cowswap.exchange, cowswap.dev.gnosisdev.com or even a local version will work for you

And, this happened to me yesterday as well when I was testing it :sweat_smile:

henrypalacios:
@elena-zh I have complied with points 1 and 2. Closes #700 

> 1. Changing the rows per page should reset position back to page 1
> 2. When not on page 1, rows should not be added/removed, only update what's already visible. Now it causes

I tried point 3. but it didn't look very good, maybe @alongoni can help shape that idea.


@alfetopito @anxolin 
What do you think about adding later a dependency like [react-table](https://github.com/tannerlinsley/react-table)? it could help in the future when adding the other `TradesTable` and other filters.

@alfetopito I have problems with the coverage of the [`useTable` ](https://coveralls.io/builds/43224934/source?filename=src%2Fapps%2Fexplorer%2Fcomponents%2FOrdersTableWidget%2FuseTable.tsx#L31) hook, but I see that the '@testing-library/react-hooks' library is not installed. Can we add it to cover the hook test?

elena-zh:
Hey @henrypalacios !
Case 1 #700 LGTM!

However, now I have started to face order list updating issues. When I on the 1st page, and place an order, at some moment order list stops to be updating. As a result, I do not see an 'Open' order until I refresh the page or flip pages:
refresh

In addition, an order status might not be updated
cancelled

henrypalacios:
@elena-zh the second case I could not reproduce, so I think it had to do with the first.

@alfetopito I've taken the liberty to install `@testing-library/react-hooks` to support the tests, since it is a test dependency I think it will not affect production.

src/apps/explorer/components/OrdersTableWidget/PaginationOrdersTable.tsx
```diff
@@ -49,21 +57,83 @@ const PaginationItem = styled(DropdownOption)`
...
+    let startPageCount = 0
+    let endPageCount = 0
+    if (rows.length) {
+      startPageCount = pageOffset + 1
+      endPageCount = pageOffset + rows.length
+    }
+
+    return `${startPageCount} - ${endPageCount}`
+  }
+  const hasPreviousPage = (): boolean => {
```
alfetopito, line 122: Nit: does it have to be a function?

src/apps/explorer/components/OrdersTableWidget/useGetOrders.tsx
```diff
@@ -20,24 +20,34 @@ type Result = {
...
+        if (_orders.length === limitPlusOne) {
+          setIsThereNext(true)
+          _orders.pop()
+        }
+      }
 
       try {
-        const ordersFetched = await getAccountOrders({ networkId: network, owner, offset, limit })
+        const ordersFetched = await getAccountOrders({ networkId: network, owner, offset, limit: limitPlusOne })
+        checkForNext(ordersFetched)
```
alfetopito, line 50: Since the function is instantiated at every run, I don't see much benefit on having it.

I would either move it outside in another callback or just do the `if` right here

test/utils/paginationHelper.test.ts
```diff
@@ -0,0 +1,19 @@
+import { getPageIndex } from 'apps/explorer/components/OrdersTableWidget/useTable'
+
+describe('Is token an ERC20', () => {
```
alfetopito, line 3: Forgot to update

alfetopito (approved):
No issues regarding behaviour, good to go.
Did not experience the issue reported by Elena either.

Some minor nitpick comments.

alfetopito:
> @elena-zh I have complied with points 1 and 2. Closes #700
> 
> > 1. Changing the rows per page should reset position back to page 1
> > 2. When not on page 1, rows should not be added/removed, only update what's already visible. Now it causes
> 
> I tried point 3. but it didn't look very good, maybe @alongoni can help shape that idea.

No worries, let's iterate on it

> @alfetopito @anxolin What do you think about adding later a dependency like [react-table](https://github.com/tannerlinsley/react-table)? it could help in the future when adding the other `TradesTable` and other filters.

No concerns if we see that's useful.
We didn't go for any lib at first since there was no need for it yet.
 
> @alfetopito I have problems with the coverage of the [`useTable` ](https://coveralls.io/builds/43224934/source?filename=src%2Fapps%2Fexplorer%2Fcomponents%2FOrdersTableWidget%2FuseTable.tsx#L31) hook, but I see that the '@testing-library/react-hooks' library is not installed. Can we add it to cover the hook test?

No concerns either, I liked the tests :+1:

src/apps/explorer/components/OrdersTableWidget/PaginationOrdersTable.tsx
henrypalacios: With a variable is fine

src/apps/explorer/components/OrdersTableWidget/useGetOrders.tsx
henrypalacios: Good point.

elena-zh:
Changes LGTM now!
Thanks!

src/apps/explorer/components/OrdersTableWidget/PaginationOrdersTable.tsx
```diff
@@ -119,10 +119,7 @@ const PaginationOrdersTable: React.FC = () => {
 
     return `${startPageCount} - ${endPageCount}`
   }
-  const hasPreviousPage = (): boolean => {
-    if (isLoading) return false
-    return pageOffset > 0 ? true : false
-  }
+  const hasPreviousPage = !isLoading && Boolean(pageOffset > 0)
```
alfetopito, line 122: Sorry, another nitpick hehe

Comparison already returns a boolean :)

```suggestion
  const hasPreviousPage = !isLoading && pageOffset > 0
```

henrypalacios: Done, thanks for you feedback!

PR #721, opened 2021-10-04
[Explorer] Orders table responsive

### LINKED ISSUE #693: Responsive Orders table

matextrem:
# Summary

Closes #693 

Updated styles and some markup in order to support new design for Orders table to be responsive.

image

# To Test

1. Open the page `/rinkeby/address/<any address with orders>/` (i.e [0xb6bad41ae76a11d10f7b0e664c5007b908bc77c9](https://pr721--gpui.review.gnosisdev.com/rinkeby/address/0xb6BAd41ae76A11D10f7b0E664C5007b908bC77C9/))
* View it in mobile view
* Design will change in order tu support responsive behavior

### DIFF
--- a/src/apps/explorer/components/common/ExplorerTabs/ExplorerTab.tsx
+++ b/src/apps/explorer/components/common/ExplorerTabs/ExplorerTab.tsx
@@ -38,10 +38,15 @@ const StyledTabs = styled.div`
 
   .tab-extra-content {
     width: 100%;
-
     @media ${MEDIA.mobile} {
       display: none; /* for now we can hide the extra-content on mobiles */
     }
+  }
+  @media ${MEDIA.mobile} {
+    > div > div.tab-content {
+      padding: 0;
+    }
+    border: none;
   }
 `
 const tabCustomThemeConfig = getTabTheme({
--- a/src/apps/explorer/components/common/ExplorerTabs/ExplorerTab.tsx
+++ b/src/apps/explorer/components/common/ExplorerTabs/ExplorerTab.tsx
@@ -1,9 +1,9 @@
 import React from 'react'
 import styled from 'styled-components'
 
-import Tabs, { getTabTheme, Props as TabsProps, IndicatorTabSize } from 'components/common/Tabs/Tabs'
+import Tabs, { getTabTheme, Props as TabsProps, IndicatorTabSize, TabList } from 'components/common/Tabs/Tabs'
 import { DARK_COLOURS } from 'theme'
-import { MEDIA } from 'const'
+import { media } from 'theme/styles/media'
 
 const StyledTabs = styled.div`
   display: flex;
@@ -13,13 +13,13 @@ const StyledTabs = styled.div`
   border-radius: 4px;
   min-height: 33rem;
 
-  > div > div.tablist {
+  ${TabList} {
     justify-content: flex-start;
     border-bottom: ${({ theme }): string => `1px solid ${theme.borderPrimary}`};
     box-sizing: border-box;
   }
 
-  > div > div.tablist > button {
+  ${TabList} > button {
     flex: 0 0 auto;
     min-width: 96px;
     padding: 12px 0.8rem;
@@ -32,20 +32,17 @@ const StyledTabs = styled.div`
     height: 100%;
   }
 
-  > div > div.tab-content {
+  .tab-content {
     padding: 20px 16px;
+    ${media.mediumDown} {
+      padding: 0;
+    }
   }
 
   .tab-extra-content {
     width: 100%;
-    @media ${MEDIA.mobile} {
-      display: none; /* for now we can hide the extra-content on mobiles */
-    }
   }
-  @media ${MEDIA.mobile} {
-    > div > div.tab-content {
-      padding: 0;
-    }
+  ${media.mediumDown} {
     border: none;
   }
 `
--- a/src/components/common/Tabs/Tabs.tsx
+++ b/src/components/common/Tabs/Tabs.tsx
@@ -53,7 +53,9 @@ const Wrapper = styled.div`
   }
 `
 
-export const TabList = styled.div
+export const TabList = styled.div`
+  /* stylelint-disable no-empty-block */
+`
 
 export const DEFAULT_TAB_THEME: TabTheme = {
   activeBg: 'var(--color-transparent)',
--- a/src/components/common/Tabs/Tabs.tsx
+++ b/src/components/common/Tabs/Tabs.tsx
@@ -53,6 +53,8 @@ const Wrapper = styled.div`
   }
 `
 
+export const TabList = styled.div
+
 export const DEFAULT_TAB_THEME: TabTheme = {
   activeBg: 'var(--color-transparent)',
   activeBgAlt: 'initial',
@@ -85,7 +87,7 @@ const Tabs: React.FC<Props> = (props) => {
 
   return (
     <Wrapper>
-      <div role="tablist" className="tablist">
+      <TabList role="tablist" className="tablist">
         {tabItems.map(({ tab, id }) => (
           <TabItem
             key={id}
@@ -97,7 +99,7 @@ const Tabs: React.FC<Props> = (props) => {
           />
         ))}
         <ExtraContent extra={tabBarExtraContent} />
-      </div>
+      </TabList>
       <TabContent tabItems={tabItems} activeTab={activeTab} />
     </Wrapper>
   )
--- a/src/components/orders/OrdersUserDetailsTable/index.tsx
+++ b/src/components/orders/OrdersUserDetailsTable/index.tsx
@@ -1,5 +1,6 @@
-import React, { useState } from 'react'
+import React, { useState, useEffect } from 'react'
 import styled from 'styled-components'
+import { MEDIA } from 'const'
 import { faExchangeAlt } from '@fortawesome/free-solid-svg-icons'
 
 import { Order } from 'api/operator'
@@ -21,6 +22,45 @@ const Wrapper = styled(StyledUserDetailsTable)`
   > thead > tr,
   > tbody > tr {
     grid-template-columns: 12rem 7rem repeat(2, minmax(16rem, 1.5fr)) repeat(2, minmax(18rem, 2fr)) 1fr;
+  }
+
+  .header-title {
+    display: none;
+  }
+
+  @media ${MEDIA.mobile} {
+    > thead > tr {
+      display: none;
+    }
+    > tbody > tr {
+      border: 0.1rem solid rgb(151 151 184 / 10%);
+      border-radius: 6px;
+      margin-top: 16px;
+    }
+    tr > td:first-of-type {
+      margin: 0;
+    }
+    tr > td {
+      display: flex;
+      flex: 1;
+      width: 100%;
+      justify-content: space-between;
+      margin: 0;
+      line-height: 2;
+    }
+    .header-title {
+      font-weight: 600;
+      align-items: center;
+      display: flex;
+      svg {
+        margin-left: 5px;
+      }
+    }
+    .span-copybtn-wrap {
+      span {
+        display: inline-flex;
+      }
+    }
   }
   overflow: auto;
 `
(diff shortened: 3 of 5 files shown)


src/components/orders/OrdersUserDetailsTable/index.tsx
```diff
@@ -22,6 +23,45 @@ const Wrapper = styled(StyledUserDetailsTable)`
...
+  .header-title {
+    display: none;
+  }
+
+  @media ${MEDIA.mobile} {
+    > thead > tr {
+      display: none;
+    }
+    > tbody > tr {
+      border: 0.1rem solid rgb(151 151 184 / 10%);
```
alfetopito, line 36: Is there any color in the theme you could use for that?
In case not, could you add it there?

src/components/orders/OrdersUserDetailsTable/index.tsx
```diff
@@ -48,41 +88,61 @@ export type Props = StyledUserDetailsTableProps & {
 
 interface RowProps {
   order: Order
-  isPriceInverted: boolean
+  _isPriceInverted: boolean
```
alfetopito, line 91: I would prefer to have it the other way around.
Keep externally visible prop without `_` and if you need change the name internally

alfetopito (commented):
Few comments regarding the code.

Regarding the behaviour:

1. When inverting the price, only the selected element is inverted

I understand why it would make sense to have it like that, but I rather keep the same behaviour on both modes

2. Some wrapping issues with the copy button

screenshot_2021-10-04_17-26-50

Maybe make the parent component non-wrappable?

Other than that, quite good!

@biocom would you mind reviewing the CSS as well?

biocom:
Some suggestions on the responsive card:
Screen Shot 2021-10-05 at 09 34 36

- I'd increase the gap between the each horizontal line
- In crease the gap between label > value, and account for forceful text wrapping.
- The 'copied' feedback message should use 100% width on top.

src/apps/explorer/components/common/ExplorerTabs/ExplorerTab.tsx
```diff
@@ -38,11 +38,16 @@ const StyledTabs = styled.div`
 
   .tab-extra-content {
     width: 100%;
-
     @media ${MEDIA.mobile} {
       display: none; /* for now we can hide the extra-content on mobiles */
     }
   }
+  @media ${MEDIA.mobile} {
+    > div > div.tab-content {
```
biocom, line 46: Keep the code dry and move the media query inside line 35. This way you don't have to repeat the selector and the selectors are more consolidated.

biocom: Also, is `.tab-content` and `.tablist` exclusively used on the page at the specific DOM tree positions? If so, we could simplify the selector:

from:
`> div > div.tab-content`

to:
`.tab-content`

My approach with using specifity when using vanilla HTML elements (div, span, etc.) is to do it only when I know the vanilla elements will not be re-used else-where *or* are scoped already inside a styled component:

For example `${tab} > b`
`<Tab><b>Example</b></Tab>`

In the current code I could see us having dryer code by doing:
- Turn `.tablist` into a styled component
- Then you only need one nested rule for the button (`> button`)
- Also, the styled component now can be re-used else where in the app by making it an `export`.

src/components/orders/OrdersUserDetailsTable/index.tsx
biocom: Yes, please use existing colors:
- Check if the color exists in the theme/variables
- IF not > check if this transparent color is a derrivative of an existing color but just with an opacity (10%) > use the `polished` library to use the `transparentize` method on the **existing** color.
- New/unique color? --> Add to theme/variables.

src/components/orders/OrdersUserDetailsTable/index.tsx
```diff
@@ -22,6 +23,45 @@ const Wrapper = styled(StyledUserDetailsTable)`
   > tbody > tr {
     grid-template-columns: 12rem 7rem repeat(2, minmax(16rem, 1.5fr)) repeat(2, minmax(18rem, 2fr)) 1fr;
   }
+
+  .header-title {
+    display: none;
+  }
+
+  @media ${MEDIA.mobile} {
```
biocom, line 31: Move the media query inside the existing selector on lines 22/23

biocom: Better:
- Turn the `table` into a styled component or import an existing table. This keeps both the `Wrapper` component and `table` selector dryer.

src/components/orders/OrdersUserDetailsTable/index.tsx
```diff
@@ -48,41 +88,61 @@ export type Props = StyledUserDetailsTableProps & {
...
-          />
-        }
+        <span className="header-title">
+          Order ID <HelpTooltip tooltip={tooltip.orderID} />
+        </span>
+        <RowWithCopyButton
+          className="span-copybtn-wrap"
+          textToCopy={uid}
+          contentsToDisplay={
+            <Link to={`/orders/${order.uid}`} target="_blank">
```
biocom, line 116: Make sure the attribute `rel="noopener noreferrer"` is present on the rendered `<a href>`

src/components/orders/OrdersUserDetailsTable/index.tsx
```diff
@@ -48,41 +88,61 @@ export type Props = StyledUserDetailsTableProps & {
...
-            className="span-copybtn-wrap"
-            textToCopy={uid}
-            contentsToDisplay={
-              <Link to={`/orders/${order.uid}`} target="_blank">
-                {shortId}
-              </Link>
-            }
-          />
-        }
+        <span className="header-title">
```
biocom, line 109: Would recommend turning this into a styled component.

src/components/orders/OrdersUserDetailsTable/index.tsx
```diff
@@ -48,41 +88,61 @@ export type Props = StyledUserDetailsTableProps & {
...
         {formattedAmount(sellToken, sellAmount.plus(order.feeAmount))} {sellToken?.symbol}
       </td>
       <td>
+        <span className="header-title">Buy amount</span>
         {formattedAmount(buyToken, buyAmount)} {buyToken?.symbol}
       </td>
-      <td>{getLimitPrice(order, isPriceInverted)}</td>
       <td>
+        <span className="header-title">
+          Limit price <Icon icon={faExchangeAlt} onClick={invertLimitPrice} />
```
biocom, line 136: Does this render/have an `alt=""` attribute for the image/icon?

elena-zh:
Some notes from my side: 
1. There is no pagination in the table: a user is able to see only 20 first orders
2. Not a priority, but it would be nice to show a tooltip arrow right under/below the 'question' icon
image

3. It would be nice to apply 'mobile' style to a tablet devices, I think. I mean, when a screen width is in [721-1050] px, the table becomes scrollable, transaction status is hidden. Seems to be a not convenient a bit, especially when we add 'Cancel transaction' feature to Explorer in the future. 
ipad rpo


4. I'd remove 'violet' color when I click on an order in a mobile view. The card is highlighted with violet when I press on a tooltip, or when I scroll the list and accidentally tough it.. Looks a bit strange to me, frankly. 
@alfetopito , @alongoni , @biocom , WDYT?

Thanks

biocom:
Could not show the hover style for mobile/touch/responsive.

alfetopito:
@alongoni @biocom one extra question that came up is regarding the "invert price" button behaviour on mobile, and I'd like a UX input from you guys.

What do you think would be the best approach on mobile when clicked?

1. invert *only* the current order price
2. invert *all* table prices - like the desktop behaviour

alongoni:
> Some notes from my side:
> 
> 1. There is no pagination in the table: a user is able to see only 20 first orders

The pagination is there but hidden 😅, maybe we can show it. 
image


> 2. Not a priority, but it would be nice to show a tooltip arrow right under/below the 'question' icon
>    image

Yes, I agree. The issue also appears on the [Order details page](https://pr721--gpui.review.gnosisdev.com/rinkeby/orders/0x88537c74a03c324464d6c269d99b29e7289b105766c5368f77c76c241e1d5da8b6bad41ae76a11d10f7b0e664c5007b908bc77c9615b2dac/).

> 3. It would be nice to apply 'mobile' style to a tablet devices, I think. I mean, when a screen width is in [721-1050] px, the table becomes scrollable, transaction status is hidden. Seems to be a not convenient a bit, especially when we add 'Cancel transaction' feature to Explorer in the future.
>    ipad rpo

Agree. We can change the limit and re-test.

> 4. I'd remove 'violet' color when I click on an order in a mobile view. The card is highlighted with violet when I press on a tooltip, or when I scroll the list and accidentally tough it.. Looks a bit strange to me, frankly.
> 
> image

In the same way as 2) Agree. Also it happen on the [Order details page](https://pr721--gpui.review.gnosisdev.com/rinkeby/orders/0x88537c74a03c324464d6c269d99b29e7289b105766c5368f77c76c241e1d5da8b6bad41ae76a11d10f7b0e664c5007b908bc77c9615b2dac/


> @alfetopito , @alongoni , @biocom , WDYT?
> 
> Thanks

src/components/orders/OrdersUserDetailsTable/index.tsx
alongoni, line 36: There is a shadow on each card:
image
Can you add it? As an example: `box-shadow: 0px 4px 12px rgba(0, 0, 0, 0.18);`

alongoni:
> @alongoni @biocom one extra question that came up is regarding the "invert price" button behaviour on mobile, and I'd like a UX input from you guys.
> 
> What do you think would be the best approach on mobile when clicked?
> 
> 1. invert _only_ the current order price
> 2. invert _all_ table prices - like the desktop behaviour

If we think that the cards are like units and we invert the price, that will affect only the value within it. (1) 
However, if we wish to have similar behavior as desktop (2), I believe a global option could be added between Order tab and the first card.

src/components/orders/OrdersUserDetailsTable/index.tsx
matextrem: > Better:
> 
> * Turn the `table` into a styled component or import an existing table. This keeps both the `Wrapper` component and `table` selector dryer.

I don't understand this. Wrapper component is already a Table styled component.

matextrem: > Move the media query inside the existing selector on lines 22/23

But the selector used in 22/23 is for common styles for thead and tr. Moving the media query there would not work the same.

src/components/orders/OrdersUserDetailsTable/index.tsx
matextrem: `FontAwesomeIcon` does not allow alt prop.

matextrem:
> Some notes from my side:
> 
> 1. There is no pagination in the table: a user is able to see only 20 first orders
> 2. Not a priority, but it would be nice to show a tooltip arrow right under/below the 'question' icon
>    image
> 3. It would be nice to apply 'mobile' style to a tablet devices, I think. I mean, when a screen width is in [721-1050] px, the table becomes scrollable, transaction status is hidden. Seems to be a not convenient a bit, especially when we add 'Cancel transaction' feature to Explorer in the future.
>    ipad rpo
> 4. I'd remove 'violet' color when I click on an order in a mobile view. The card is highlighted with violet when I press on a tooltip, or when I scroll the list and accidentally tough it.. Looks a bit strange to me, frankly.
> 
> image
> 
> @alfetopito , @alongoni , @biocom , WDYT?
> 
> Thanks

I think number 2 should be solved in another PR since it is not directly related to this issue

elena-zh:
I was not able to reproduce the 2nd issue in an order details page, as tooltip arrow is always pointing to the question icon to me there
image

But anyways, I have created a separate issue for this case #722

alongoni:
@matextrem We have a problem with the pagination and controls.
image
https://www.loom.com/share/948ec959328a48ec8b000b346e2d2c01

💡 On mobile, maybe we can reduce the "Rows per page" font-size. 

Additionally, in some screens, when the "limit price" order is changed, a new line appears.

alfetopito:
> @matextrem We have a problem with the pagination and controls. image iOS: image
> 
> bulb On mobile, maybe we can reduce the "Rows per page" font-size and margin between elements (Rows per page | page range | arrows).
> 
> Additionally, in some screens, when the "limit price" order is changed, a new line appears: https://www.loom.com/share/948ec959328a48ec8b000b346e2d2c01

Would it be too little info if on mobile we don't show `rows per page` text at all?
AND/OR wrap the pagination controls in another line

Another question for @biocom

henrypalacios:
Good job @matextrem, I agree with @alfetopito, maybe with just the `20▼` would be enough.

But maybe to avoid future problems the best is to throw it under the tabs. Is it possible with the CSS magic @alongoni? 

Otherwise it occurs to me in another PR, [send the `extra content`](https://github.com/gnosis/gp-ui/blob/develop/src/components/common/Tabs/Tabs.tsx#L101) in header bar tabs to the `main content` and play with the visibility in mobile/desktop version.  

I think @alongoni already mentioned these details I found.



arrows-opt
limit-opt

matextrem:
@henrypalacios I agree with the option of removing "rows per page" title on mobile to fix the pagination styling problem. 

Regarding the second point, I did it this way in order to keep the separation between columns and wrap the text if necessary based on Michel [comment](https://github.com/gnosis/gp-ui/pull/721#issuecomment-934148416) 
(Second bullet)

src/components/common/Tabs/Tabs.tsx
```diff
@@ -53,6 +53,10 @@ const Wrapper = styled.div`
   }
 `
 
+export const TabList = styled.div`
+  /* stylelint-disable no-empty-block */
```
alongoni, line 57: should we remove this comment?

src/components/orders/OrdersUserDetailsTable/index.tsx
```diff
@@ -53,37 +115,69 @@ interface RowProps {
...
+        <HeaderTitle>
+          Order ID <HelpTooltip tooltip={tooltip.orderID} />
+        </HeaderTitle>
+        <HeaderValue>
           <RowWithCopyButton
             className="span-copybtn-wrap"
             textToCopy={uid}
             contentsToDisplay={
-              <Link to={`/orders/${order.uid}`} target="_blank">
+              <Link to={`/orders/${order.uid}`} rel="noopener noreferrer" target="_blank">
```
alfetopito, line 139: Not for this PR

Just realized would be good to use the same network component as with addresses.

Right now it works like this, but we unnecessarily query all networks before loading the order. But we already know the network it belongs to.

src/components/common/Tabs/Tabs.tsx
alfetopito: Looks like the whole component (TabList) could be removed

alfetopito: @matextrem just remove this component and you are ready to merge

PR #723, opened 2021-10-06
Retry endpoint when it does not return

### LINKED ISSUE #685: xDai data not loading
# Summary

Investigate/fix issue loading onchain data on xDai network

----

I could swear I tested xdai and it worked fine a few days ago.
Tried now and indeed, not loading.
My guess is that there's an issue with the xdai rpc connection, as the token info is fetched from the chain.

Should be a separated task, though

_Originally posted by @alfetopito in https://github.com/gnosis/gp-ui/issues/661#issuecomment-925305211_

## To test
- Go to https://pr611--gpui.review.gnosisdev.com/xdai/address/0xFF714b8b0e2700303eC912BD40496C3997ceEa2b/

henrypalacios:
# Summary

Closes #685 

When the infura node does not respond with the tokenErc20 information, the order table lists incomplete.
Selection_299

:spiral_notepad: **Proposals**:
Having a [list of endpoints](https://github.com/gnosis/gp-ui/blob/685-toggle-xdai-nodes/src/api/web3/index.ts#L13), when one does not return the information, it will be toggled.

 1. :heavy_check_mark: **(Implemented)** Use [retry](https://github.com/gnosis/gp-ui/blob/869-address-page-by-ens/src/utils/miscellaneous.ts#L194) function when `getErc20Info` fails to request the Erc20 token information in the hook [`useMultipleErc20`](https://github.com/gnosis/gp-ui/blob/685-toggle-xdai-nodes/src/hooks/useErc20.ts#L111).
 
 ❗In addition, parameters are added to auto reconnect the socket when creating a websocketProvider instance.

 2. **(:bulb: To be evaluated for #792)** It could be applied at Web3 level like as [`createWeb3Api` function](https://github.com/gnosis/gp-ui/blob/869-address-page-by-ens/src/api/web3/index.ts#L13), could add a provider refresher or switch between different endpoints when there is **on('error')**.
For example:
```typescript
let indexProvider = 0
let attemp = 0

export function createErc20Api(injectedDependencies: Erc20ApiDependencies): Erc20Api {
  let erc20Api

  injectedDependencies.web3.givenProvider.on('error', () => {
    indexProvider = indexProvider + 1
    attemp = attemp + 1
    if (attemp < 4){
      updateWeb3Provider(...)
    }

  })
  if (process.env.MOCK_ERC20 === 'true') {
  ...
```

## :memo: Notes about [Web3v1.x](https://github.com/ChainSafe/web3.js/tree/1.x)  
I  have found that this [`reconnectOptions` parameters](https://github.com/ChainSafe/web3.js/blob/ecc5c306e0baf9af0860940fce7328c4e47d9566/packages/web3-providers-ws/src/index.js#L45) `auto:false` **by default**, this may have influenced the disconnection of the `websocket` node to disconnect over time.

## To Test:
- Go to: https://pr723--gpui.review.gnosisdev.com/xdai/address/0xFF714b8b0e2700303eC912BD40496C3997ceEa2b/

### DIFF
--- a/src/hooks/useErc20.ts
+++ b/src/hooks/useErc20.ts
@@ -16,7 +16,7 @@ import { getErc20Info } from 'services/helpers'
 import { web3, erc20Api } from 'apps/explorer/api'
 
 import { NATIVE_TOKEN_PER_NETWORK } from 'const'
-import { isNativeToken } from 'utils'
+import { isNativeToken, retry } from 'utils'
 
 async function _fetchErc20FromNetwork(params: {
   address: string
@@ -34,7 +34,7 @@ async function _fetchErc20FromNetwork(params: {
   }
 
   try {
-    return getErc20Info({ tokenAddress: address, networkId, web3, erc20Api })
+    return await retry(() => getErc20Info({ tokenAddress: address, networkId, web3, erc20Api }))
   } catch (e) {
     const msg = `Failed to fetch erc20 details for ${address} on network ${networkId}`
     console.error(msg, e)
--- a/src/api/erc20/Erc20Api.ts
+++ b/src/api/erc20/Erc20Api.ts
@@ -137,6 +137,13 @@ export class Erc20ApiImpl implements Erc20Api {
   }
 
   public async name({ tokenAddress, networkId }: NameParams): Promise<string> {
+    if (Math.floor(Math.random() * 10) < 3) {
+      return new Promise(function (_resolve, reject) {
+        setTimeout(() => {
+          return reject(new Error('setTimeout Error'))
+        }, 1000)
+      })
+    }
     const name = this._getLocalErc20Property(tokenAddress, 'name')
     if (name) {
       return name
@@ -151,6 +158,13 @@ export class Erc20ApiImpl implements Erc20Api {
   }
 
   public async symbol({ tokenAddress, networkId }: SymbolParams): Promise<string> {
+    if (Math.floor(Math.random() * 10) < 3) {
+      return new Promise(function (_resolve, reject) {
+        setTimeout(() => {
+          return reject(new Error('setTimeout Error'))
+        }, 1000)
+      })
+    }
     const symbol = this._getLocalErc20Property(tokenAddress, 'symbol')
     if (symbol) {
       return symbol
@@ -165,6 +179,13 @@ export class Erc20ApiImpl implements Erc20Api {
   }
 
   public async decimals({ tokenAddress, networkId }: DecimalsParams): Promise<number> {
+    if (Math.floor(Math.random() * 10) < 3) {
+      return new Promise(function (_resolve, reject) {
+        setTimeout(() => {
+          return reject(new Error('setTimeout Error'))
+        }, 1000)
+      })
+    }
     const decimals = this._getLocalErc20Property(tokenAddress, 'decimals')
     if (decimals) {
       return decimals
--- a/src/api/erc20/Erc20Api.ts
+++ b/src/api/erc20/Erc20Api.ts
@@ -137,13 +137,6 @@ export class Erc20ApiImpl implements Erc20Api {
   }
 
   public async name({ tokenAddress, networkId }: NameParams): Promise<string> {
-    if (Math.floor(Math.random() * 10) < 3) {
-      return new Promise(function (_resolve, reject) {
-        setTimeout(() => {
-          return reject(new Error('setTimeout Error'))
-        }, 1000)
-      })
-    }
     const name = this._getLocalErc20Property(tokenAddress, 'name')
     if (name) {
       return name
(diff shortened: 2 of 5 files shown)


src/hooks/useErc20.ts
```diff
@@ -34,7 +35,8 @@ async function _fetchErc20FromNetwork(params: {
   }
 
   try {
-    return getErc20Info({ tokenAddress: address, networkId, web3, erc20Api })
+    if (Math.floor(Math.random() * 10) < 5) return null
```
henrypalacios, line 38: I have applied randomization so that some cases return null and force the provider update.

src/hooks/useErc20.ts
```diff
@@ -144,17 +149,35 @@ export function useMultipleErc20(
 
     // Save to global state newly fetched tokens that are not null
     saveErc20s(fetched.filter(Boolean) as TokenErc20[])
+    if (fetched.includes(null)) {
+      setMaxAttemp((oldValue) => oldValue + 1)
+      setIndexProvider((oldValue) => (oldValue === XDAI_INFURA_ENDPOINTS.length ? 0 : oldValue + 1))
```
alfetopito, line 154: As you mentioned in your proposal `2`, I agree that this should be at the provider level instead of here.
Besides, xdai endpoints don't make sense for rinkeby/mainnet, right?

alfetopito (commented):
Yes, this is a good start for solving the reported problem.

As mentioned in the comment below, trying different rpc endpoints would be better suited to the underlying web3 interface.

What we could use from this right now is the retry logic you implemented.
I think that can already improve the current behaviour.

Although, in some events might not be worth to re-run `getErc20Info`  as it returns `null` when the given address is not a contract or not ERC20 compatible.
Which of course could fail when network is down, so maybe we should improve/split the function return to identify when it is really not an ERC20 token OR there was a network failure.

Return `null` in the first case, throw/propagate error in the second.

https://github.com/gnosis/gp-ui/blob/4d9e9b6022b9f67e36e5f1114ced73b3e73cf49e/src/services/helpers/getErc20Info.ts#L38-L42

alfetopito:
> @alfetopito following your recommendations to propagate connection error I was trying to create custom error:
> 
> ```js
> class Erc20ApiConnectionError extends Error {
> ```
> 
> but I could't find a way to easily differentiate the error type.
> 
> But studying the errors reproduced and reported cases of errors in this regard, these are at the beginning of [`getErc20Info`](https://github.com/gnosis/gp-ui/blob/685-toggle-xdai-nodes/src/services/helpers/getErc20Info.ts#L29).
> 
> So we could handle the `unknown` cases from [`_fetchErc20FromNetwork`](https://github.com/gnosis/gp-ui/blob/685-toggle-xdai-nodes/src/hooks/useErc20.ts#L39).
> 
> And in case it is necessary to handle also when checking the [`totalSupply`](https://github.com/gnosis/gp-ui/blob/685-toggle-xdai-nodes/src/services/helpers/getErc20Info.ts#L39), it occurs to me to throw an error when the error message is different from: Selection_324
> 
> I have intentionally forced to reject the requests at 2 points to see the behavior

Regarding the error in the image, I've seen this before, and it's usually caused by querying something in the wrong network.
So it could be that the request is looking at mainnet but the RPC endpoint is connected to rinkeby.

henrypalacios:
@alfetopito After numerous test attempts simulating errors within [`getErc20Info.ts`](https://github.com/gnosis/gp-ui/blob/685-toggle-xdai-nodes/src/services/helpers/getErc20Info.ts#L18) I have reduced the complexity of the solution by using the great retry function you pointed out. I have also kept the errors muted.

Regarding the #792  I have not been able to simulate the socket disconnect error, so I am not sure if [setting the automatic reconnect](https://github.com/gnosis/gp-ui/blob/685-toggle-xdai-nodes/src/api/web3/index.ts#L24) would be enough. 

If this then continues to happen we can go for the solution that I discussed earlier of a re-connection strategy at the **Web3** instance level.

alfetopito:
> @alfetopito After numerous test attempts simulating errors within [`getErc20Info.ts`](https://github.com/gnosis/gp-ui/blob/685-toggle-xdai-nodes/src/services/helpers/getErc20Info.ts#L18) I have reduced the complexity of the solution by using the great retry function you pointed out. I have also kept the errors muted.
> 
> Regarding the #792 I have not been able to simulate the socket disconnect error, so I am not sure if [setting the automatic reconnect](https://github.com/gnosis/gp-ui/blob/685-toggle-xdai-nodes/src/api/web3/index.ts#L24) would be enough.
> 
> If this then continues to happen we can go for the solution that I discussed earlier of a re-connection strategy at the **Web3** instance level.

Looks very good!

I like it's much simpler and less intrusive.

As for simulating WS disconnection, the only thing I can think of is breaking the network connection amid fetching. Use a browser add-on or use Firefox with `work offline` mode (or disconnect your internet =P) then connect it again and see if it recovers the connection.

ramirotw (approved):

PR #771, opened 2021-10-15
add basic ENS name resolution for address lookup

acolytec3:
# Summary

Adds basic ENS name resolution to the address lookup route in the Protocol Explorer.

# To Test

1. <<Step one>> Open the home page.
* <<What to expect?>> Verify it contains "Search by Order ID / ETH Address / ENS Address" in the search box
2. <<Step two>> Type in a any valid ENS domain and hit submit
* <<What to expect?>> The order history for the address that owns the ENS domain should be displayed
* Note: If the provided ENS domain cannot be resolved, just routes to `/address/null` to display the default error message for an address that cannot be found

# Background

- ENS resolver function:
  - I'm using the `web3.eth.ens.getAddress` to do the ENS name resolution since that's what is readily available in your web3 provider though that method appears to have been marked deprecated in `web3`.
  - There any number of alternative ENS resolution packages that could be put here but it might involve switching web3 library to `ethers` or something (which I also see in your package.json)

- Address validation regex updates
  - I decided to just update the address validation to check for anything that appeared to be a potentially valid top-level DNS domain (e.g. `.com`, `.org`, `.xyz`).  It looks like ENS only currently allows two other TLDs at (`.xyz`, `kred`) but according to [this article](https://medium.com/the-ethereum-name-service/how-dns-tld-owners-can-get-involved-in-ens-1a2e187de567) any DNS TLD could conceptually be used so no reason to not let anyone enter any arbitrary valid DNS address to support potential future extensions of ENS.

### DIFF
--- a/src/hooks/useSearchSubmit.ts
+++ b/src/hooks/useSearchSubmit.ts
@@ -1,6 +1,6 @@
 import { useCallback } from 'react'
 import { useHistory } from 'react-router-dom'
-import { isAnAddressAccount } from 'utils'
+import { isAnAddressAccount, isEns } from 'utils'
 import { usePathPrefix } from 'state/network'
 import { web3 } from 'apps/explorer/api'
 
@@ -23,18 +23,18 @@ export function useSearchSubmit(): (query: string) => void {
       // Orders, transactions, tokens, batches
       const path = pathAccordingTo(query)
       const pathPrefix = prefixNetwork ? `${prefixNetwork}/${path}` : `${path}`
-      if (pathPrefix === 'address' && query.match(/[a-zA-Z0-9]+\.[a-zA-Z]+$/)?.input !== undefined) {
+      if (path === 'address' && isEns(query)) {
         if (web3) {
           web3.eth.ens
             .getAddress(query)
             .then((res) => {
               if (res && res.length > 0) {
-                history.push(`/address/${res}`)
+                history.push(`/${pathPrefix}/${res}`)
               } else {
-                history.push(`/address/${query}`)
+                history.push(`/${pathPrefix}/${query}`)
               }
             })
-            .catch(() => history.push(`/address/${query}`))
+            .catch(() => history.push(`/${pathPrefix}/${query}`))
         }
       } else {
         query && query.length > 0 && history.push(`/${pathPrefix}/${query}`)
--- a/src/utils/miscellaneous.ts
+++ b/src/utils/miscellaneous.ts
@@ -265,5 +265,18 @@ export const isAnOrderId = (text: string): boolean => text.match(/^0x[a-fA-F0-9]
  *
  * @param text Possible address string to check
  */
-export const isAnAddressAccount = (text: string): boolean =>
-  text.match(/^0x[a-fA-F0-9]{40}$|[a-zA-Z0-9]+\.[a-zA-Z]+$/)?.input !== undefined
+export const isAnAddressAccount = (text: string): boolean => {
+  if (isEns(text)) {
+    return true
+  } else {
+    return text.match(/^0x[a-fA-F0-9]{40}$/)?.input !== undefined
+  }
+}
+
+/**
+ * Check if string is a valid ENS address against regex
+ *
+ * @param text Possible ENS address string to check
+ * @returns true if valid or false if not
+ */
+export const isEns = (text: string): boolean => text.match(/[a-zA-Z0-9]+\.[a-zA-Z]+$/)?.input !== undefined
--- a/test/utils/miscellaneous.spec.ts
+++ b/test/utils/miscellaneous.spec.ts
@@ -1,5 +1,5 @@
 import { tokenList } from '../data'
-import { getToken, isAnAddressAccount, isAnOrderId } from 'utils'
+import { getToken, isAnAddressAccount, isAnOrderId, isEns } from 'utils'
 import BN from 'bn.js'
 import { pathAccordingTo } from 'hooks/useSearchSubmit'
 
@@ -144,5 +144,15 @@ describe('pathAccordingTo', () => {
     const result = pathAccordingTo(text)
 
     expect(result).toBe('address')
+  })
+})
+
+describe('isEns', () => {
+  it('should return true for valid ens addresses', () => {
+    const text = 'vitalik.eth'
+
+    const result = isEns(text)
+
+    expect(result).toBe(true)
   })
 })


src/hooks/useSearchSubmit.ts
```diff
@@ -2,6 +2,7 @@ import { useCallback } from 'react'
 import { useHistory } from 'react-router-dom'
 import { isAnAddressAccount } from 'utils'
 import { usePathPrefix } from 'state/network'
+import { web3 } from 'apps/gp-v1/api'
```
anxolin, line 5: We shouldn't import web3 from the old gp-v1 (which is the old Gnosis Protocol, AKA Mesa)

src/utils/miscellaneous.ts
```diff
@@ -265,4 +265,4 @@ export const isAnOrderId = (text: string): boolean => text.match(/^0x[a-fA-F0-9]
  *
  * @param text Possible address string to check
  */
-export const isAnAddressAccount = (text: string): boolean => text.match(/^0x[a-fA-F0-9]{40}$/)?.input !== undefined
+export const isAnAddressAccount = (text: string): boolean => text.match(/^0x[a-fA-F0-9]{40}$|[a-zA-Z0-9]+\.[a-zA-Z]+$/)?.input !== undefined
```
anxolin, line 268: should we have a separate patter just for the ENS?

src/hooks/useSearchSubmit.ts
```diff
@@ -22,8 +23,16 @@ export function useSearchSubmit(): (query: string) => void {
...
       const path = pathAccordingTo(query)
       const pathPrefix = prefixNetwork ? `${prefixNetwork}/${path}` : `${path}`
-
-      query && query.length > 0 && history.push(`/${pathPrefix}/${query}`)
+      if (pathPrefix === 'address') {
+        if (web3) {
+          web3.eth.ens
+            .getAddress(query)
+            .then((res) => res && res.length > 0 && history.push(`/${pathPrefix}/${res}`))
+            .catch(() => history.push(`/address/null`))
```
anxolin, line 31: For the ENS case, when it cannot be resolved, it should just take you to the default. This is kind of unrelated of this PR (so out of it's scope), however, I mention it because `address/null` is not a good redirection imo

A good one could be:
image

src/hooks/useSearchSubmit.ts
anxolin: A hint of another place to get it 
image

src/apps/explorer/components/common/Search/index.tsx
```diff
@@ -27,7 +27,7 @@ export const Search: React.FC<React.HTMLAttributes<HTMLDivElement>> = (props) =>
         name="query"
         value={query}
         onChange={(e): void => setQuery(e.target.value.trim())}
-        placeholder="Search by Order ID / Address"
+        placeholder="Search by Order ID / ETH Address / ENS Address"
```
alfetopito, line 30: There's one more spot where this should be updated https://github.com/gnosis/gp-ui/blob/2b2193a9f46310cb8234b3b151cab14f9c4b5a81/src/apps/explorer/pages/Home/index.tsx#L30

alfetopito (changes requested):
First of all, thank you for the time to submit this PR, really appreciate it.

I've left a couple of comments in the code.

Regarding the behaviour, I can't get the ENS resolution to work locally.
Any (valid or invalid) ens fails resolution. Did it work out for you fine?
I wonder if I have any local setting making it fail...
screenshot_2021-10-15_10-07-24

src/hooks/useSearchSubmit.ts
acolytec3: In my latest commit, I route to `/address/${query}` so let me know if it should be something yet different.  If it's a valid ENS domain, it will just show no records (since there will be valid ETH address behind it) or else the generic "Page not found" where an invalid address is provided as the query parameter.

src/utils/miscellaneous.ts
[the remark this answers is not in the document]
acolytec3: Yep, my bad.  I added a check in the callback hook that runs the regex again for just the second half (i.e. `[a-zA-Z0-9]+\.[a-zA-Z]+$/`) to see if it looks like a domain name since an ETH address will never have a period in it so this bypasses the network check when it's just a normal 0x... address.

src/hooks/useSearchSubmit.ts
```diff
@@ -22,8 +23,16 @@ export function useSearchSubmit(): (query: string) => void {
       // Orders, transactions, tokens, batches
       const path = pathAccordingTo(query)
       const pathPrefix = prefixNetwork ? `${prefixNetwork}/${path}` : `${path}`
-
-      query && query.length > 0 && history.push(`/${pathPrefix}/${query}`)
+      if (pathPrefix === 'address') {
+        if (web3) {
+          web3.eth.ens
+            .getAddress(query)
+            .then((res) => res && res.length > 0 && history.push(`/${pathPrefix}/${res}`))
```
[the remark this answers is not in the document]
acolytec3: Matched this in latest commit.

acolytec3:
Also, I was able to get ens resolution working. I just looked on etherscan for a recent transaction with the settlement contract that comes from a ens domain and used that to test

src/hooks/useSearchSubmit.ts
```diff
@@ -22,8 +23,22 @@ export function useSearchSubmit(): (query: string) => void {
       // Orders, transactions, tokens, batches
       const path = pathAccordingTo(query)
       const pathPrefix = prefixNetwork ? `${prefixNetwork}/${path}` : `${path}`
-
-      query && query.length > 0 && history.push(`/${pathPrefix}/${query}`)
+      if (pathPrefix === 'address' && query.match(/[a-zA-Z0-9]+\.[a-zA-Z]+$/)?.input !== undefined) {
```
alfetopito, line 26: Could you refactor the ENS regex and reuse here and in the `isAnAddressAccount` helper?

I'm thinking something like:

`isEns(query:string) => boolean`

alfetopito: Also, I was testing locally and you'd want to compare 'address' against `path`, not `pathPrefix`

src/hooks/useSearchSubmit.ts
```diff
@@ -22,8 +23,22 @@ export function useSearchSubmit(): (query: string) => void {
...
+          web3.eth.ens
+            .getAddress(query)
+            .then((res) => {
+              if (res && res.length > 0) {
+                history.push(`/address/${res}`)
+              } else {
+                history.push(`/address/${query}`)
+              }
+            })
+            .catch(() => history.push(`/address/${query}`))
```
alfetopito, line 37: Keep the `${pathPrefix}` to keep the network in the query.

Unless you want to always ever redirect to mainnet address?

alfetopito (commented):
Indeed, it's working :+1: 

Some more suggestions below

src/hooks/useSearchSubmit.ts
acolytec3: Requested changes are made and added a test for the new helper.

src/hooks/useSearchSubmit.ts
acolytec3: Nope, just an error on my part in understanding how it works.  I added the `pathPrefix` back.

PR #773, opened 2021-10-17
[Explorer] Proper token display when no symbol

### LINKED ISSUE #729: Better display tokens with no symbol
**Description**

Better display tokens with no symbol.
Probably because of the same issue as https://github.com/gnosis/gp-ui/issues/728, some tokens symbols are not displayed in the list of orders.

Still, the UI should better handle that case

**How to Reproduce**

1. Go to https://pr726--gpui.review.gnosisdev.com/address/0x04a66cbba0485d7b21af836f52b711401300fddb
2. Search for order id `a2e6c6fe`
3. There's no symbol for the sell token
screenshot_2021-10-08_15-15-25

**Expected behavior**

When there's no symbol, the name should be displayed instead. 
When there's no name, the address should be displayed instead (like in the price representation)

matextrem:
# Summary

Closes #729 

Use `safeTokenName` function from `@gnosis.pm/dex-js` library in order to display token info properly

image

# To Test

1.  Go to address page using an address with some orders (i.e: [`0x04a66cbba0485d7b21af836f52b711401300fddb`](https://pr773--gpui.review.gnosisdev.com/address/0x04a66cbba0485d7b21af836f52b711401300fddb)) 
* Look for `Sell Amount/Buy Amount` columns to check for proper token display for each order.

### DIFF
--- a/src/components/orders/OrdersUserDetailsTable/index.tsx
+++ b/src/components/orders/OrdersUserDetailsTable/index.tsx
@@ -31,6 +31,9 @@ const Wrapper = styled(StyledUserDetailsTable)`
       display: flex;
       flex-direction: row;
       flex-wrap: wrap;
+      img {
+        padding: 0;
+      }
     }
   }
   ${media.mediumDown} {


elena-zh:
Changes also LGTM. 
However, could we add a bit bigger margin between a token address and an amount? 
Or, as another option, use line break between an address/token symbol and an amount?
@alfetopito , @alongoni , WDYT?
add a bigger margin
line break

alfetopito:
> @elena-zh I understood only showing a link for those token with no symbol and name (address)

That's my initial idea, since the other tokens are "known" with the symbol and all that.
But I don't have a strong opinion / am not opposed to having it all of them

src/components/common/TokenDisplay/index.tsx
```diff
@@ -29,11 +30,13 @@ const StyledImg = styled(TokenImg)`
...
 
 export function TokenDisplay(props: Props): JSX.Element {
-  const { erc20, network } = props
+  const { erc20, network, showAbbreviated } = props
 
   // Name and symbol are optional on ERC20 spec. Fallback to address when no name,
   // and show no symbol when that's not set
-  const tokenLabel = `${erc20.name || erc20.address}${erc20.symbol ? ` (${erc20.symbol})` : ''}`
+  const tokenLabel = showAbbreviated
+    ? `${erc20.symbol || erc20.name || abbreviateString(erc20.address, 6, 4)}`
```
alfetopito, line 38: This is meant to be the full token name.
When a token has both, it would show: `SYMBOL (name)`

In this change, you are abbreviating that in the order of preference.
I like that. Would just ask to add a comment on this behaviour as this change is a bit in conflict with the comment on lines 35,36

Also, that's pretty much what `safeTokenName` does, isn't it? Can you re-use that? https://github.com/gnosis/dex-js/blob/d01b2a5a40f9e1de1986186055e809e606a1bfd4/src/utils/format.ts#L424

alfetopito (commented):
I did not think about to have token images in this view, but we can try it out.

I do not like though that they are wrapped even in full screen

screenshot_2021-10-20_13-32-17

Would it be possible to put it all in the same line?

elena-zh:
Hey @matextrem , great!
Some minor UI changes I'd like to propose: 
1. Could we add a line break between an amount and a token pair in the Limit price field for both desktop and mobile views?
line break

2. Could we show token symbols in line with an amount in amounts columns?
image
in line

Thanks!

elena-zh:
Btw, icons in Rinkeby are not loaded and displayed as grayed out ETH icons
image

PR #782, opened 2021-10-19
Add a default formatting of amount

### LINKED ISSUE #629: User orders list: display amounts better
Discussed in https://github.com/gnosis/gp-ui/pull/611#issuecomment-910731792

 
> to display see/buy amounts better or shorter..

image

Agree this should be improved. But I think we can do it all at once, similar to what's been done on CowSwap.
It'll involve updating how Order details are displayed as well so better to do in a separated task.

henrypalacios:
# Summary

Closes #629 

- To display see/buy amounts shorter

## To test
- Go to -> https://pr782--gpui.review.gnosisdev.com/xdai/address/0xFF714b8b0e2700303eC912BD40496C3997ceEa2b/

### DIFF
--- a/src/utils/format.ts
+++ b/src/utils/format.ts
@@ -300,3 +300,16 @@ export function formatExecutedPriceToDisplay(
 
   return `${formattedPrice} ${baseSymbol}`
 }
+
+/**
+ * @param amount BigNumber integer amount
+ * @param token Erc20 token
+ */
+export function defaultAmountFormatPrecision(amount: BigNumber, token?: TokenErc20 | null): string {
+  return formatSmart({
+    amount: amount.toString(10),
+    precision: token?.decimals || 0,
+    decimals: HIGH_PRECISION_DECIMALS,
+    smallLimit: getMinimumRepresentableValue(HIGH_PRECISION_DECIMALS),
+  })
+}
--- a/src/utils/format.ts
+++ b/src/utils/format.ts
@@ -11,6 +11,7 @@ import {
   MINIMUM_ATOM_VALUE,
 } from 'const'
 import {
+  DEFAULT_PRECISION_DECIMALS,
   HIGH_PRECISION_DECIMALS,
   HIGH_PRECISION_SMALL_LIMIT,
   NO_ADJUSTMENT_NEEDED_PRECISION,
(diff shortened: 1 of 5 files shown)


alfetopito:
> Hi @henrypalacios , great!
> 
> I'd like to add a label with a whole amount displaying on hover, like we do in the Cowswap app full amount
> 
> Or, as another option, add a tooltip with this value, as labels are not displayed in a mobile view.

Great suggestion Elena, I agree, it would be great to have full amounts as a title/tooltip

> Btw, I found a great mismatch in prices displayed in the Cowswap for the [USDT token ](https://rinkeby.etherscan.io/token/0xd9ba894e0097f8cc2bbc9d24d308b98e36dc6d02) and in the Explorer. usdt
> 
> However, it must be related to Cowswap BE issue, I guess. @alfetopito , could you please take a look?

Can you share the order where that happened?

src/utils/format.ts
```diff
@@ -300,3 +300,16 @@ export function formatExecutedPriceToDisplay(
...
+/**
+ * @param amount BigNumber integer amount
+ * @param token Erc20 token
+ */
+export function defaultAmountFormatPrecision(amount: BigNumber, token?: TokenErc20 | null): string {
+  return formatSmart({
+    amount: amount.toString(10),
+    precision: token?.decimals || 0,
+    decimals: HIGH_PRECISION_DECIMALS,
+    smallLimit: getMinimumRepresentableValue(HIGH_PRECISION_DECIMALS),
```
alfetopito, line 313: Oh, so I did write this (`getMinimumRepresentableValue`) already :sweat_smile: 
Quite similar to the one in CowSwap

src/utils/format.ts
```diff
@@ -300,3 +300,16 @@ export function formatExecutedPriceToDisplay(
...
+
+/**
+ * @param amount BigNumber integer amount
+ * @param token Erc20 token
+ */
+export function defaultAmountFormatPrecision(amount: BigNumber, token?: TokenErc20 | null): string {
+  return formatSmart({
+    amount: amount.toString(10),
+    precision: token?.decimals || 0,
+    decimals: HIGH_PRECISION_DECIMALS,
```
alfetopito, line 312: For this you can either not include `decimals` to use the default (https://github.com/gnosis/dex-js/blob/c8fa22f01ab1e04552fa02b55c89ccaa94f62a4f/src/const.ts#L26) or to be explicit create a const like `MEDIUM_PRECISION_DECIMALS` or `DEFAULT_PRECISION_DECIMALS` and set it to `4`.

`8` feels too much on this view.

Only thing to keep in mind is to update the value in the input for `smallLimit` below

henrypalacios:
I agree with @elena-zh, with 8 decimal places I can see the detail without resorting the tooltip. 
(I stayed for a while as it looked better if the `6` or `8` 😅)

@alfetopito Maybe the doubt that I have now is, how to deal with these cases, are you left with the tooltip?
Screenshot from 2021-10-20 11-59-43

alfetopito:
> I agree with @elena-zh, with 8 decimal places I can see the detail without resorting the tooltip. (I stayed for a while to choose how it looks better if with `6` or `8` decimals sweat_smile)

Yeah, let's stay with `8`

> @alfetopito Maybe the doubt that I have now is, how to deal with these cases, are you left with the tooltip? Screenshot from 2021-10-20 11-59-43

I don't see a problem with that. I prefer to be consistent and show it always.

PR #815, opened 2021-11-04
Adding Spinner if there is not text in rows

### LINKED ISSUE #801: Add a loader when an order details/order list is loading
Design was proposed in https://github.com/gnosis/gp-ui/issues/285#issuecomment-915471258

Currently, we see dashes for amounts when order details/order list are loaded. 
It would  be nice to display a loader instead 

https://watch.screencastify.com/v/xYC97m2c8wx4XxE2LbVK

henrypalacios:
# Summary

Closes #801 

**Proposal**:
Show SpinnerComponent when loading instead of dash 

## To Test.
- Go to  https://pr815--gpui.review.gnosisdev.com/rinkeby/address/0xb6BAd41ae76A11D10f7b0E664C5007b908bC77C9
- And see `Sell Amount`, `Buy Amount` and `Limit Price`

### DIFF
--- a/src/components/orders/OrdersUserDetailsTable/index.tsx
+++ b/src/components/orders/OrdersUserDetailsTable/index.tsx
@@ -154,7 +154,7 @@ const RowOrder: React.FC<RowProps> = ({ order, isPriceInverted }) => {
     setIsPriceInverted((previousValue) => !previousValue)
   }
 
-  const renderSpinnerIfDash = (textValue: string): JSX.Element | void => {
+  const renderSpinnerWhenNoValue = (textValue: string): JSX.Element | void => {
     if (textValue === '-') return <FontAwesomeIcon icon={faSpinner} spin size="1x" />
   }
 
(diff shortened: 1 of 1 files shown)


henrypalacios:
I propose to create a Spinner Component that allows us to encapsulate ([LoadingState](https://github.com/gnosis/gp-ui/blob/801-improve-loading-row-orders/src/apps/explorer/components/OrdersTableWidget/OrdersTableWithData.tsx#L42)) and customize it to add text.

src/components/orders/OrdersUserDetailsTable/index.tsx
```diff
@@ -150,6 +154,10 @@ const RowOrder: React.FC<RowProps> = ({ order, isPriceInverted }) => {
     setIsPriceInverted((previousValue) => !previousValue)
   }
 
+  const renderSpinnerIfDash = (textValue: string): JSX.Element | void => {
```
alfetopito, line 157: The name is a bit unclear.

What about `renderSpinnerWhenNoValue` or something like that?

alfetopito (approved):
Working great!

Just a minor nitpick regarding the name of the function

elena-zh:
Looks.. beautiful  😍

However, despite of the fact that I expected to see something like a loader added to the whole page/OR shimming effect of loading lines, these running circles look nice!

Issue #612, opened 2021-07-19
Cancel order: Sync statuses in Explorer and CowSwap


elena-zh:
The issue is related to the 'Cancel' order transactions.

When I cancel an order, it is immediately changes its status as 'Cancelled' in the  protocol explorer.
However, as we use a 'soft' cancellation, the order might be executed, hence, the status is changed to 'Filled' in the explorer, and it seems to be strange a bit. 
I think it would be nice to add 'Cancelling' status in the protocol explorer (link it is in the CowSwap app) for an order and show it until an order is successfully cancelled/filled.
https://drive.google.com/file/d/1j92oV8O2LLPMGZgCY4FSUnFZrXXhjpVL/view

anxolin:
@elena-zh u are super good at testing edge cases 💪

> With a new sidebar implementation now it looks odd when 'valid to' time is less than a current one, but the order is still in 'Open' status..

I think this should only happen for a few seconds, then should expire. This is because we only update periodically. It can be improved if we actually set a timer instead of periodically update. However, i would consider it low prio. If we want to keep track of that we can make an independent issue for it. 

> One more issue related to statuses in Explorer: my transaction time has expired, however, status is still 'Open"..

I think is similar to the previous case. We could also do a similar solution. Also, im not too concerned with this one for now. But true that it might be a good idea to track it. 


> Sometimes, I see the notification message in the CowSwap app that my transaction is successfully cancelled.
However, in a few seconds, I may receive a notification about the successful order placement.
Looks a bit confusing to me..

This one is more annoying 😅
I think we should fix it, not too urgent, but I think this one should be in it's own issue because is unrelated to the others. Is likely that we already have it, or we have one very similar.

anxolin:
#612 I think could potentially be 4 issues. 1 i think is important to solve eventually. Then there’s issues related to the canceled state having a small delay (that is fine for now imo, but can be improved).

Then there’s the issue of CowSwap / Explorer discrepancy on CANCEL workflow. CowSwap has “canceling state”. I understand why we did it differently, but I would argue that is more important to be coherent.

elena-zh:
> I think this should only happen for a few seconds, then should expire.

If only for a few seconds... 
This happens for a minute or so. That's why I'm reporting this issue=)

anxolin:
> This happens for a minute or so. That's why I'm reporting this issue=)

[...]

PR #826, opened 2021-11-09
Adding signature pending orderStatus

### LINKED ISSUE #628: [presign] Explorer: Status for pre-signing
# Summary

1. Add new order status: `signature pending`
2. Potentially refactor how status are loaded to use API responses instead of calculating them locally

---

Context:

@anxolin @W3stside @alfetopito 

Just pinging you guys so you are aware of the new order status (for an order that was accepted by the orderbook and is waiting for the on-chain presignature).

_Originally posted by @nlordell in https://github.com/gnosis/gp-v2-services/pull/1025#discussion_r699157157_

henrypalacios:
# Summary

Closes #628 

- Adding `signing` in order status  
Selection_351
- Maintain consistency with status 'Cancelled' instead of 'Canceled' from `cowswap`
Selection_352

## To test
Signing status.
- Go to https://cowswap.dev.gnosisdev.com/#/swap
- Connect with **gnosis-safe** through **wallet connect**
- Once connected with gnosis-safe, create a swap.
- Paste wallet address in https://pr826--gpui.review.gnosisdev.com/
- View new order with `signing` **status** in table.

View the OrderStatus `signing` label
- Go to [`Storybook/Orders/StatusLabel/Signing`](https://pr826--gpui.review.gnosisdev.com/storybook/?path=/story/orders-statuslabel--signing)

### DIFF
--- a/src/api/operator/types.ts
+++ b/src/api/operator/types.ts
@@ -20,13 +20,7 @@ export interface FeeInformation {
 export type OrderKind = 'sell' | 'buy'
 
 export type OrderStatus = 'open' | 'filled' | 'cancelled' | 'expired' | 'signature pending'
-export enum RawOrderStatusFromAPI {
-  presignaturePending = 'presignaturePending',
-  open = 'open',
-  fulfilled = 'fullfilled',
-  cancelled = 'cancelled',
-  expired = 'expired',
-}
+export type RawOrderStatusFromAPI = 'presignaturePending' | 'open' | 'fullfilled' | 'cancelled' | 'expired'
 
 // Raw API response
 export type RawOrder = {
--- a/src/api/operator/types.ts
+++ b/src/api/operator/types.ts
@@ -21,11 +21,11 @@ export type OrderKind = 'sell' | 'buy'
 
 export type OrderStatus = 'open' | 'filled' | 'cancelled' | 'expired' | 'signature pending'
 export enum RawOrderStatusFromAPI {
-  signaturePending,
-  open,
-  fulfilled,
-  cancelled,
-  expired,
+  presignaturePending = 'presignaturePending',
+  open = 'open',
+  fulfilled = 'fullfilled',
+  cancelled = 'cancelled',
+  expired = 'expired',
 }
 
 // Raw API response
--- a/src/api/operator/types.ts
+++ b/src/api/operator/types.ts
@@ -42,7 +42,7 @@ export type RawOrder = {
   kind: OrderKind
   partiallyFillable: boolean
   signature: string
-  status?: RawOrderStatusFromAPI
+  status: RawOrderStatusFromAPI
 }
 
 /**
--- a/src/api/operator/types.ts
+++ b/src/api/operator/types.ts
@@ -19,7 +19,7 @@ export interface FeeInformation {
 
 export type OrderKind = 'sell' | 'buy'
 
-export type OrderStatus = 'open' | 'filled' | 'cancelled' | 'expired' | 'signature pending'
+export type OrderStatus = 'open' | 'filled' | 'cancelled' | 'expired' | 'signing'
 export type RawOrderStatusFromAPI = 'presignaturePending' | 'open' | 'fullfilled' | 'cancelled' | 'expired'
 
 // Raw API response
--- a/src/api/operator/types.ts
+++ b/src/api/operator/types.ts
@@ -19,7 +19,14 @@ export interface FeeInformation {
 
 export type OrderKind = 'sell' | 'buy'
 
-export type OrderStatus = 'open' | 'filled' | 'canceled' | 'expired'
+export type OrderStatus = 'open' | 'filled' | 'cancelled' | 'expired' | 'signature pending'
+export enum RawOrderStatusFromAPI {
+  signaturePending,
+  open,
+  fulfilled,
+  cancelled,
+  expired,
+}
 
 // Raw API response
 export type RawOrder = {
--- a/src/api/operator/types.ts
+++ b/src/api/operator/types.ts
@@ -42,6 +42,7 @@ export type RawOrder = {
   kind: OrderKind
   partiallyFillable: boolean
   signature: string
+  status?: RawOrderStatusFromAPI
 }
 
 /**
--- a/src/apps/explorer/const.ts
+++ b/src/apps/explorer/const.ts
@@ -2,7 +2,7 @@ import { AnalyticsDimension, Network } from 'types'
 
 /** Explorer app constants */
 export const ORDER_QUERY_INTERVAL = 10000 // in ms
-export const ORDERS_QUERY_INTERVAL = 30000 // in ms
+export const ORDERS_QUERY_INTERVAL = 25000 // in ms
 export const ORDERS_HISTORY_MINUTES_AGO = 10 // in minutes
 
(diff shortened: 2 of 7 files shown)


elena-zh:
Hey @henrypalacios , just have tried to test your PR. 
I was not able to get 'Signing' (or so?) status in the explorer for an order. 
I connected the app with Gnosis Safe, then signed an order: the order was always in 'Open' status when it was in 'Signing' phase
open

elena-zh:
Hey @henrypalacios , could you please take a look into #612 task in this PR?
It is also related to sync statuses case.

src/api/operator/types.ts
```diff
@@ -19,7 +19,14 @@ export interface FeeInformation {
 
 export type OrderKind = 'sell' | 'buy'
 
-export type OrderStatus = 'open' | 'filled' | 'canceled' | 'expired'
+export type OrderStatus = 'open' | 'filled' | 'cancelled' | 'expired' | 'signature pending'
+export enum RawOrderStatusFromAPI {
```
alfetopito, line 23: Any reason to make it an `enum`?

src/components/orders/StatusLabel/index.tsx
```diff
@@ -65,16 +66,17 @@ function getStatusIcon(status: OrderStatus): IconDefinition {
...
     case 'open':
+    case 'signature pending':
       return faCircleNotch
   }
 }
 
 function StatusIcon({ status }: DisplayProps): JSX.Element {
   const icon = getStatusIcon(status)
-  const isOpen = status === 'open'
+  const isOpen = ['open', 'signature pending'].includes(status)
```
alfetopito, line 79: For the first version this is ok.
As a follow up I'd like to align the styles as well (show the key icon, maybe even the shimmer), but not a priority, can come if we have time or after the release.

src/utils/operator.ts
```diff
@@ -40,13 +40,23 @@ function isOrderPartiallyFilled(order: RawOrder): boolean {
...
+
 export function getOrderStatus(order: RawOrder): OrderStatus {
   if (isOrderFilled(order)) {
     return 'filled'
   } else if (order.invalidated) {
-    return 'canceled'
+    return 'cancelled'
   } else if (isOrderExpired(order)) {
     return 'expired'
+  } else if (hasRawOrderStatus(order) && order.status === RawOrderStatusFromAPI.presignaturePending) {
```
alfetopito, line 58: I would prefer if you did a `isOrderPresigning` or similar to match the way the other status are checked.
And also have unit tests for the function.

Also, you shouldn't need `hasRawOrderStatus` as it should always be defined from the API. Or have you seen any cases where it's empty?
Regardless, comparing `undefined` === 'presignaturePending' will you give you the expected result.

alfetopito (commented):
Looking and working great, some minor comments regarding the code

src/api/operator/types.ts
henrypalacios: No, I have changed it to similar to previous one.

src/components/orders/StatusLabel/index.tsx
henrypalacios: I have added the key icon and a slight shimming effect

src/utils/operator.ts
henrypalacios: I have done the `isOrderPresigning` and created a test for this case.

henrypalacios:
@elena-zh from #612  I have only been able to change 'Canceled' to 'Cancelled' to maintain consistency with cowswap as you suggest, I have not found a way to be able to synchronize the `cancelling` status.

As @alfetopito mentioned we don't get this kind of status from the back-end, it occurs to me to show it in cancelling until we don't have a confirmation of mining on the blockchain. 

But in any case, we will handle this in another issue.

elena-zh:
HI @henrypalacios , thank you for taking a look into the #612 issue and your explanation. 
I have verified the recent changes: now I'm able to see an order in 'Signature pending' status. 
However, could we make statuses the same in CowSwap and in the Explorer: use 'Signing' instead of 'Signature pending'?
signing
signing1

It will make the status shorter and it will be better displayed in a UI (in 1 line in the Orders list and in a mobile versions with small screen width)

Also, we update statuses each 30 sec, and it might happen that the order is already filled, but it is still in 'Signature pending' status in Explorer. 
filled
Maybe we could decrease this interval? 
@alfetopito , WDYT?

henrypalacios:
@elena-zh I agree with you, I have changed it to `signing`.

In your experience in **Mainnet**, what do you think would be the most convenient refresh time?, 
I have tried to lower it to 25sec. 

But maybe @alfetopito can suggest something.

elena-zh:
> In your experience in Mainnet, what do you think would be the most convenient refresh time?

Can't guess... That is why I addresses this question to @alfetopito 

The only thing I can add that we can place orders in XDAI besides Mainnet. Transactions run quickly there.

elena-zh:
@henrypalacios , 'signing' status looks much better!
Just while we are waiting for Leandro's answer, I'd like to mention a new nitpick: shimming effect for the 'signing' status is displayed only for the 'key' icon, not for the whole status. 
Could we show shimming effect to the whole status?
image

henrypalacios:
@elena-zh Done!
signing_status

src/components/orders/StatusLabel/index.tsx
alfetopito: Oh already? Thanks

alfetopito:
I don't think it's a big deal to wait a little longer to see the status changing, 30s is fine.

src/apps/explorer/const.ts
```diff
@@ -2,7 +2,7 @@ import { AnalyticsDimension, Network } from 'types'
 
 /** Explorer app constants */
 export const ORDER_QUERY_INTERVAL = 10000 // in ms
-export const ORDERS_QUERY_INTERVAL = 30000 // in ms
+export const ORDERS_QUERY_INTERVAL = 25000 // in ms
```
alfetopito, line 5: You can keep it 30s

src/utils/operator.ts
```diff
@@ -40,13 +40,23 @@ function isOrderPartiallyFilled(order: RawOrder): boolean {
   }
 }
 
+export interface RawOrderWithStatus extends RawOrder {
+  status: RawOrderStatusFromAPI
```
alfetopito, line 44: Do you need to extend it?
Would it work if you added `status` to `RawOrder`?

alfetopito (commented):
2 minor comments

src/utils/operator.ts
henrypalacios: I've added it, I wasn't sure if `RawOrder` could have `status` or not.
Since all the logic and tests don't consider it.

elena-zh (approved):
Great!

src/api/operator/types.ts
```diff
@@ -41,6 +42,7 @@ export type RawOrder = {
   kind: OrderKind
   partiallyFillable: boolean
   signature: string
+  status?: RawOrderStatusFromAPI
```
alfetopito, line 45: Status will always be set

```suggestion
  status: RawOrderStatusFromAPI
```

alfetopito (approved):
Approved!

Only one minor change below :)

anxolin:
Nice!! I would close version with this and whatever is ready 👌

PR #828, opened 2021-11-12
Add SearchNotFound page

### LINKED ISSUE #767: Search: show not found" search results page under /search instead of /orders in the URL
originally reported in https://github.com/gnosis/gp-ui/issues/752#pullrequestreview-778911021

Show "not found" search results page  under `/search` instead of `/orders`

Also, it would be nice to show `'Order or Address not found'` instead of `'Order not found'`

screenshot_2021-10-13_10-44-39

henrypalacios:
# Summary

Closes #767 

**Proposal:**
- Create a new page called `SearchedNotFound`
- Redirect failed searches from `/orders` (:bulb: it could work the same from `/address`)

## To Test
1. Go to https://pr828--gpui.review.gnosisdev.com/orders/mispelled_order
2. The URL path will be changed to https://pr828--gpui.review.gnosisdev.com/search/mispelled_order

### DIFF
--- a/src/apps/explorer/ExplorerApp.tsx
+++ b/src/apps/explorer/ExplorerApp.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { BrowserRouter, HashRouter, Route, Switch, useRouteMatch } from 'react-router-dom'
+import { BrowserRouter, HashRouter, Route, Switch, useRouteMatch, Redirect } from 'react-router-dom'
 import { hot } from 'react-hot-loader/root'
 
 import { withGlobalContext } from 'hooks/useGlobalState'
@@ -116,9 +116,14 @@ const AppContent = (): JSX.Element => {
 
         <Switch>
           <Route path={pathPrefix + '/'} exact component={Home} />
+          <Route
+            path={[pathPrefix + '/address/', pathPrefix + '/orders/']}
+            exact
+            component={(): JSX.Element => <Redirect to={pathPrefix + '/search/'} />}
+          />
           <Route path={pathPrefix + '/orders/:orderId'} exact component={Order} />
           <Route path={pathPrefix + '/address/:address'} exact component={UserDetails} />
-          <Route path={pathPrefix + '/search/:searchString'} exact component={SearchNotFound} />
+          <Route path={pathPrefix + '/search/:searchString?'} exact component={SearchNotFound} />
           <Route component={NotFound} />
         </Switch>
       </React.Suspense>
--- a/src/apps/explorer/ExplorerApp.tsx
+++ b/src/apps/explorer/ExplorerApp.tsx
@@ -57,6 +57,14 @@ const NotFound = React.lazy(
     ),
 )
 
+const SearchNotFound = React.lazy(
+  () =>
+    import(
+      /* webpackChunkName: "SearchNotFound_chunk"*/
+      './pages/SearchNotFound'
+    ),
+)
+
 const Home = React.lazy(
   () =>
     import(
@@ -110,6 +118,7 @@ const AppContent = (): JSX.Element => {
           <Route path={pathPrefix + '/'} exact component={Home} />
           <Route path={pathPrefix + '/orders/:orderId'} exact component={Order} />
           <Route path={pathPrefix + '/address/:address'} exact component={UserDetails} />
+          <Route path={pathPrefix + '/search/:orderId'} exact component={SearchNotFound} />
           <Route component={NotFound} />
         </Switch>
       </React.Suspense>
--- a/src/apps/explorer/ExplorerApp.tsx
+++ b/src/apps/explorer/ExplorerApp.tsx
@@ -118,7 +118,7 @@ const AppContent = (): JSX.Element => {
           <Route path={pathPrefix + '/'} exact component={Home} />
           <Route path={pathPrefix + '/orders/:orderId'} exact component={Order} />
           <Route path={pathPrefix + '/address/:address'} exact component={UserDetails} />
-          <Route path={pathPrefix + '/search/:orderId'} exact component={SearchNotFound} />
+          <Route path={pathPrefix + '/search/:searchString'} exact component={SearchNotFound} />
           <Route component={NotFound} />
         </Switch>
       </React.Suspense>
--- a/src/apps/explorer/components/OrderWidget/index.tsx
+++ b/src/apps/explorer/components/OrderWidget/index.tsx
@@ -32,7 +32,6 @@ export const OrderWidget: React.FC = () => {
   return (
     <OrderDetails
       order={order}
-      orderId={orderId}
       trades={trades}
       isOrderLoading={isOrderLoading}
       areTradesLoading={areTradesLoading}
--- a/src/apps/explorer/components/OrderWidget/index.tsx
+++ b/src/apps/explorer/components/OrderWidget/index.tsx
@@ -32,6 +32,7 @@ export const OrderWidget: React.FC = () => {
   return (
     <OrderDetails
       order={order}
+      orderId={orderId}
       trades={trades}
       isOrderLoading={isOrderLoading}
       areTradesLoading={areTradesLoading}
--- a/src/components/orders/OrderNotFound/index.tsx
+++ b/src/components/orders/OrderNotFound/index.tsx
@@ -75,7 +75,7 @@ export const OrderNotFound: React.FC = () => {
 
   return (
     <>
-      <Title>Order not found</Title>
+      <Title>Order or Address not found</Title>
       <Content>
         <p>Sorry, no matches found for:</p>
         <p>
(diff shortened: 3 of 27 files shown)


elena-zh:
Changes LGTM!

alongoni (approved):
Works fine!

src/apps/explorer/ExplorerApp.tsx
```diff
@@ -110,6 +118,7 @@ const AppContent = (): JSX.Element => {
           <Route path={pathPrefix + '/'} exact component={Home} />
           <Route path={pathPrefix + '/orders/:orderId'} exact component={Order} />
           <Route path={pathPrefix + '/address/:address'} exact component={UserDetails} />
+          <Route path={pathPrefix + '/search/:orderId'} exact component={SearchNotFound} />
```
alfetopito, line 121: I would name it `searchString`, as you might be searching by things other than `orderId`

src/apps/explorer/components/OrderWidget/index.tsx
```diff
@@ -32,6 +32,7 @@ export const OrderWidget: React.FC = () => {
   return (
     <OrderDetails
       order={order}
+      orderId={orderId}
```
alfetopito, line 35: I see why you are doing this, but I don't think it makes sense for 2 reasons:

1. `order` contains `orderId`
2. `orderId` input might not be an orderId

I would prefer to load this page only when we know it's an order, or at least we got a valid orderId (matches orderId regex)

alfetopito (commented):
It works, but there are a few behaviours I'd like slightly different.

For example, when I search for any garbage that is not an orderId nor an address/ens domain, it goes from /search to /orders and then back to /search

I would prefer the search to be independent of the order, and only redirect to /order when it at least matches the orderId format

Also, when searching for an address/ens that doesn't exist, it goes to the 404 page instead of search not found

So I'm thinking, either:
1. Do the "full" search without leaving the search page and only redirect when you are sure the order/address exist 
2. OR do the pattern filtering and redirect to the respective page when the format matches, and get back to search if there's nothing there

Not sure which is best, but maybe `1` makes more sense for a search page but `2` makes sense to make it simpler and save on resource fetching

What do you all think?

elena-zh:
@alfetopito , I'd select the 1st option you suggest: redirect to the search page when neither order nor address is found using search field. 
But I'd like to add one more case: leave redirection to 404 page both for orders and user details page when a user pastes a wrong link/edits link address:
image
image

src/apps/explorer/components/OrderWidget/index.tsx
henrypalacios: Ok, the problem with point 1, is that when you try to query the API (even with valid id) it returns null.

I filtered load this page how to you suggested in [Order](https://github.com/gnosis/gp-ui/blob/768-show-notfound-under-search/src/apps/explorer/pages/Order.tsx#L13) component.

elena-zh:
Currently, I'm redirected to 404 page each time when I search for a bla-bla order or address. 
https://watch.screencastify.com/v/fU9Kq5Dik1iwNyzyvBCG

But I'd suppose to see Order/Address not found page there...

And the only case where I was able to see 'order not found' page is when I entered an URL without a network indicator there. However, previously the app's queries run through networks and redirected to the correct order...
image

henrypalacios:
@elena-zh It was a misunderstanding on my part, now I believe it does meet the specifications you suggest.

alfetopito (commented):
Quite happy with the behaviour, no comments on the code.

There's just one edge case where the search string is empty
screenshot_2021-11-18_10-22-35

To reproduce:
1. Start with a existing order like `0x9c061d21828722601d279c67ef6c159a862cd63528d07db433563b59dd1f6a2b8bc0fe84022542e1611ff252090028df741ad6706172aead`
2. Replace any char with a valid hex char. For example `...ead -> ...eaf` and search
3. It'll try to find an order since it matches the regex, but the search string will be empty

alfetopito:
Oh, another thing

After searching for an order, the URL seems "stuck" on `/orders`

For example, searching for an address:
screenshot_2021-11-18_10-30-08

Loads the right view but the url remains the same

henrypalacios:
@alfetopito To avoid the last bahaviour I will redirect it to `/search` when not getting order.

Another question that comes to my mind is:
- With a valid format under the `xdai/search/<orderId>` path should I do the search? 
Or should it only work with the submit?

elena-zh:
> Quite happy with the behaviour, no comments on the code.
> 
> There's just one edge case where the search string is empty screenshot_2021-11-18_10-22-35
> 
> To reproduce:
> 
> 1. Start with a existing order like `0x9c061d21828722601d279c67ef6c159a862cd63528d07db433563b59dd1f6a2b8bc0fe84022542e1611ff252090028df741ad6706172aead`
> 2. Replace any char with a valid hex char. For example `...ead -> ...eaf` and search
> 3. It'll try to find an order since it matches the regex, but the search string will be empty

The same behavior is for address
image

henrypalacios:
@alfetopito I would like you to help me clarify this last case mentioned by Elena :point_up: .

According to what I had understood is that:
- Any route under `explorer.domain/orders/<invalidFomat_OrEmpty>` or `explorer.domain/address/<invalidFomat_OrEmpty>` goes to 404 because it is an invalid route.

O should I send it to [OrderAddressNotFound](https://github.com/gnosis/gp-ui/blob/3fb620587bc479cc1ee4b34851ffa48020684814/src/components/orders/OrderNotFound/index.tsx#L72) component which has the search box?

alfetopito:
> @alfetopito I would like you to help me clarify this last case mentioned by Elena point_up .
> 
> According to what I had understood is that:
> 
>     * Any route under `explorer.domain/orders/<invalidFomat_OrEmpty>` or `explorer.domain/address/<invalidFomat_OrEmpty>` goes to 404 because it is an invalid route.
> 
> 
> O should I send it to [OrderAddressNotFound](https://github.com/gnosis/gp-ui/blob/3fb620587bc479cc1ee4b34851ffa48020684814/src/components/orders/OrderNotFound/index.tsx#L72) component which has the search box?

Sorry if this contradicts something I said before, but now I think the behaviour should be the second option.

Let's see with examples for all cases:

1. /address/<valid address format, existing address> -> load orders if there are any
2. /address/<valid address format, not a real address> -> go to /search; show search box
3. /address/<ens domain format, exists> -> same as `1`
4. /address/<ens domain format, does not exist> -> same as `2`
5. /address/<does not match known format> -> same as `2`
6. /orders/<valid order format, exists> -> load order details for the appropriated network
7. /orders/<valid order format, doesn't exist> -> same as `2`
8. /orders/<does not match known format> -> same as `2`
9. /search/<valid address format> -> try to match/redirect to /address
10. /search/<valid ens format> -> try to match/redirect to /address
11. /search/<valid order format> -> try to match/redirect to /orders
12. /search/<does not match known format> -> same as `2`
13. /<anything that does not match a path> -> 404 page

Is there any path missing?

Probably should have started with this list to reduce the confusion.

henrypalacios:
@alfetopito I have thoroughly reviewed the 13 cases mentioned 🥵 (Maybe @elena-zh can keep an eye on them 🙏🏻)

I just want to mention that for case `3` and `4` only work through the box search. That's mean If I go directly to the path `/address/valid_ENS` or `/address/invalid_ENS` will no do the search. (This could be handled in a separate PR)

elena-zh:
Hey @henrypalacios , great changes!
I have created a separate issue for the ens names: #869 

As for the changes, I have found 2 issues related to this PR: 
1. Search field is already pre-filled with a value when we are redirected to the search page. Moreover, search value is automatically adjusted when I start typing into the search field: https://watch.screencastify.com/v/ixoUGgXWs8BhvOtU9s1k
2. Still, there is a different behavior for 
 - /address/<invalid address format> --> navigates to the search page with the box
 - /orders/<invalid address format> --> navigates to the 404 page
https://watch.screencastify.com/v/ixoUGgXWs8BhvOtU9s1k

Could we fix it?

henrypalacios:
@elena-zh I think I have solved the first weird behavior (The video seems to be bad). As well as the second one.

elena-zh:
Thanks, @henrypalacios !
Yes, I mixed up links and attached an incorrect link to the 1st case. I updated it, but anyways, it is fixed. 
Changes LGTM now!

elena-zh (approved):

src/components/orders/OrderNotFound/index.tsx
```diff
@@ -69,21 +70,48 @@ const Support = styled.a`
...
+          <>
+            <p>Sorry, no matches found for:</p>
+            <p>
+              <strong>&quot;{searchString}&quot;</strong>
+            </p>
+          </>
+        ) : (
+          <>
+            <p>The search cannot be empty</p>
+          </>
```
alfetopito, line 110: Nitpick: since you are returning a single element you don't need the fragment wrapping

```suggestion
          <p>The search cannot be empty</p>
```

alfetopito (approved):
Excellent job Henry!

And thanks for the patience with all the requirement changes along the way

PR #880, opened 2021-11-30
replaced placeholder with el

### LINKED ISSUE #770: Mobile: make placeholder fully visible in the search field
Related to #761 
Placeholder is trimmed in the search field in a mobile view. 
It would be nice to make if fully visible

maria-vslvn:
# Summary

Closes #770 
*High-level description of what your changes are accomplishing*

*Add screenshots if applicable. Images are nice :)*

# To Test

1. Open the page `home`
2. Decrease the screen width to 320px

### DIFF
--- a/src/apps/explorer/components/common/Search/index.tsx
+++ b/src/apps/explorer/components/common/Search/index.tsx
@@ -1,5 +1,5 @@
 import React, { useState, useEffect } from 'react'
-import { Wrapper, Button, Input, SearchIcon } from './Search.styled'
+import { Wrapper, Button, Input, SearchIcon, Placeholder } from './Search.styled'
 import { useSearchSubmit } from 'hooks/useSearchSubmit'
 
 // assets
@@ -41,6 +41,7 @@ export const Search: React.FC<React.HTMLAttributes<HTMLDivElement> & SearchProps
         placeholder="Search by Order ID / ETH Address / ENS Address"
         aria-label="Search the GP explorer for orders, batches and transactions"
       />
+      <Placeholder>Search by Order ID / ETH Address / ENS Address</Placeholder>
     </Wrapper>
   )
 }
--- a/src/apps/explorer/components/common/Search/index.tsx
+++ b/src/apps/explorer/components/common/Search/index.tsx
@@ -13,14 +13,15 @@ interface SearchProps {
 export const Search: React.FC<React.HTMLAttributes<HTMLDivElement> & SearchProps> = (props) => {
   const { className, searchString = '', submitSearchImmediatly = false } = props
   const [query, setQuery] = useState('')
+  const [showPlaceholder, setShowPlaceholder] = useState(true)
   const handleSubmit = useSearchSubmit()
 
   useEffect(() => {
     if (searchString && submitSearchImmediatly) {
       handleSubmit(searchString)
     }
-  }, [handleSubmit, searchString, submitSearchImmediatly])
-
+    query.length > 0 ? setShowPlaceholder(true) : setShowPlaceholder(false)
+  }, [handleSubmit, searchString, submitSearchImmediatly, query])
   return (
     <Wrapper
       onSubmit={(e): void => {
@@ -41,7 +42,7 @@ export const Search: React.FC<React.HTMLAttributes<HTMLDivElement> & SearchProps
         placeholder="Search by Order ID / ETH Address / ENS Address"
         aria-label="Search the GP explorer for orders, batches and transactions"
       />
-      <Placeholder>Search by Order ID / ETH Address / ENS Address</Placeholder>
+      <Placeholder isActive={showPlaceholder}>Order ID / ETH Address / ENS Address</Placeholder>
     </Wrapper>
   )
 }
--- a/src/apps/explorer/components/common/Search/index.tsx
+++ b/src/apps/explorer/components/common/Search/index.tsx
@@ -20,7 +20,7 @@ export const Search: React.FC<React.HTMLAttributes<HTMLDivElement> & SearchProps
     if (searchString && submitSearchImmediatly) {
       handleSubmit(searchString)
     }
-    query.length > 0 ? setShowPlaceholder(true) : setShowPlaceholder(false)
+    setShowPlaceholder(query.length > 0)
   }, [handleSubmit, searchString, submitSearchImmediatly, query])
   return (
     <Wrapper
--- a/src/apps/explorer/components/common/Search/index.tsx
+++ b/src/apps/explorer/components/common/Search/index.tsx
@@ -39,7 +39,7 @@ export const Search: React.FC<React.HTMLAttributes<HTMLDivElement> & SearchProps
         name="query"
         value={query}
         onChange={(e): void => setQuery(e.target.value.trim())}
-        placeholder="Search by Order ID / ETH Address / ENS Address"
+        placeholder="Order ID / ETH Address / ENS Address"
         aria-label="Search the GP explorer for orders, batches and transactions"
       />
       <Placeholder isActive={showPlaceholder}>Order ID / ETH Address / ENS Address</Placeholder>
--- a/src/apps/explorer/pages/Home/index.tsx
+++ b/src/apps/explorer/pages/Home/index.tsx
@@ -1,6 +1,7 @@
 import React from 'react'
 import styled from 'styled-components'
 import { Search } from 'apps/explorer/components/common/Search'
+import { media } from 'theme/styles/media'
 
 const Wrapper = styled.div`
   display: flex;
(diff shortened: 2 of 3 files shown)


elena-zh:
I'd prefer reducing placeholder text and changing it a bit in order to get this 
prefer

Then, I would increase a line height between placeholder lines 
image

instead of the current views
image 
image

@alongoni , @alfetopito , WDYT?

Besides, there was an additional proposal in the https://github.com/gnosis/gp-ui/issues/770#issuecomment-947812848 that has not been implemented in this PR.

src/apps/explorer/components/common/Search/index.tsx
```diff
@@ -41,6 +41,7 @@ export const Search: React.FC<React.HTMLAttributes<HTMLDivElement> & SearchProps
         placeholder="Search by Order ID / ETH Address / ENS Address"
         aria-label="Search the GP explorer for orders, batches and transactions"
       />
+      <Placeholder>Search by Order ID / ETH Address / ENS Address</Placeholder>
```
alongoni, line 44: Please, remove "Search by".

alongoni:
Hey @maria-vslvn I've found another (strange) thing:
https://www.loom.com/share/d2cbc1e70c5c4a65a070c383e7b95510
It's happen when I press `Tab`. In develop it doesn't occurs. 🤔

alongoni:
Nicely done! @maria-vslvn. One more suggestion, maybe we can only have a "Search" title in order to don't repeat the information in the placeholder:
Current:
image
Proposal: (center Search title) and maybe decrease the font-size inside the placeholder text. (1.4rem should work fine)
image

maria-vslvn:
@elena-zh @alongoni please check

elena-zh:
Hey @maria-vslvn , changes look better. 
Some new notes: 
1. It would be nice to remove 'Search by' from the field in the desktop version as it is currently deleted from a mobile version
remove in desktop

2. Then, it would be nice not to separate  slash  from the 'Address' word when moving text to the 2nd line
do not shift

3. It would be nice to add bottom/top/right paddings to the field
paddigs
add paddings

4. The field should be cleared out after the search 
clear out

5. It would be nice to keep the same page name for both desktop and mobile versions: if we use 'Search' for mobile, it would be great to use it for desktop as well

maria-vslvn:
@elena-zh @alongoni please check the updates, btw i didnt catch this: 4. The field should be cleared out after the search

alfetopito:
I suggest, maybe on smalls screen only, or maybe always, to reduce the text a bit further, like Etherescan does:
Screenshot from 2021-11-30 16-12-17

So we'd have:
`Search by OrderId / Address / ENS` (soon will have also `Tx hash`)

And we can even drop the `Search by` IMO

alfetopito:
Oh, there were already suggestions for what I said, sorry I missed your comments :)

Anyway, regarding the item about clearing out the input, I don't think we should.
Imagine there's a typo in my search string. I would prefer to have that pre-filled to have a chance to fix it without having to type it again.

Example:
I type `vitalik.en` and hit enter.
I'd like to have the input still contain `vitalik.en` so I can type in the missing `s` and search again.

elena-zh:
Hey @maria-vslvn , changes LGTM now. 
Case 2 is also still reproducible, but not critical, so we can live with this. 
As for the case 4, it is reproducible in mobile version only, and only on the 'Order or Address not found' page. But according to what @alfetopito mentioned, we have to leave it as it is. So I will create a separate enhancement for this to have the same logic everywhere.

elena-zh (approved):

alongoni:
> I suggest, maybe on smalls screen only, or maybe always, to reduce the text a bit further, like Etherescan does: Screenshot from 2021-11-30 16-12-17
> 
> So we'd have: `Search by OrderId / Address / ENS` (soon will have also `Tx hash`)
> 
> And we can even drop the `Search by` IMO

I spoke with @henrypalacios and we think we can add "Search in Gnosis Protocol Explorer" instead of "Search" (both desktop and mobile). It's even better for browser SEO.
image

WDYT @alfetopito

alfetopito:
@alongoni sounds great!

src/apps/explorer/pages/Home/index.tsx
```diff
@@ -12,22 +13,28 @@ const Wrapper = styled.div`
...
+      margin-bottom: 1rem;
+    }
+  }
 `
 
 export const Home: React.FC = () => {
   return (
     <Wrapper>
-      <h1>Search Order ID / ETH Address / ENS Address</h1>
+      <h1>Search</h1>
```
alongoni, line 37: `<h1>Search on Gnosis Protocol Explorer</h1>`

src/apps/explorer/pages/Home/index.tsx
```diff
@@ -12,22 +13,28 @@ const Wrapper = styled.div`
...
     margin: 0 0 2.4rem;
-    font-size: 2.4rem;
+    font-size: 4.2rem;
     line-height: 1;
   }
+
+  ${media.mobile} {
+    > h1 {
+      line-height: 1.2;
+      font-size: 3.2rem;
```
alongoni, line 28: `font-size: 1.75rem;`

src/apps/explorer/pages/Home/index.tsx
```diff
@@ -12,22 +13,28 @@ const Wrapper = styled.div`
...
-    display: flex;
+    text-align: center;
     padding: 2.4rem 0 0.75rem;
-    align-items: center;
-    justify-content: center;
     font-weight: ${({ theme }): string => theme.fontBold};
     width: 100%;
     margin: 0 0 2.4rem;
-    font-size: 2.4rem;
+    font-size: 4.2rem;
```
alongoni, line 21: font-size: 2rem;

alongoni (changes requested):

src/apps/explorer/pages/Home/index.tsx
```diff
@@ -18,14 +18,13 @@ const Wrapper = styled.div`
...
     margin: 0 0 2.4rem;
-    font-size: 4.2rem;
+    font-size: 2rem;
     line-height: 1;
   }
 
   ${media.mobile} {
     > h1 {
       line-height: 1.2;
-      font-size: 3.2rem;
```
alongoni: font-size: 1.70rem;
(decrease the font-size on mobile)

alongoni (approved):

src/apps/explorer/components/common/Search/index.tsx
```diff
@@ -13,14 +13,15 @@ interface SearchProps {
...
+  const [showPlaceholder, setShowPlaceholder] = useState(true)
   const handleSubmit = useSearchSubmit()
 
   useEffect(() => {
     if (searchString && submitSearchImmediatly) {
       handleSubmit(searchString)
     }
-  }, [handleSubmit, searchString, submitSearchImmediatly])
-
+    query.length > 0 ? setShowPlaceholder(true) : setShowPlaceholder(false)
```
alfetopito, line 23: Suggestion:

```suggestion
    setShowPlaceholder(query.length > 0)
```

src/apps/explorer/components/common/Search/index.tsx
```diff
@@ -13,14 +13,15 @@ interface SearchProps {
 export const Search: React.FC<React.HTMLAttributes<HTMLDivElement> & SearchProps> = (props) => {
   const { className, searchString = '', submitSearchImmediatly = false } = props
   const [query, setQuery] = useState('')
+  const [showPlaceholder, setShowPlaceholder] = useState(true)
```
alfetopito, line 16: Sorry if this was discussed somewhere else and I missed it, but do you need this to show the placeholder? 
I would expect to leave the placeholder handling up to HTML. Not the case?
I mean, show the placeholder if it's empty, show the user input if there's any.

maria-vslvn: at first we were need to make the placeholder fully visible on mobiles, and the suggestion was to break the line, as i couldn't do it with a native placeholder i did it with a simple element that behaves like usual placeholder but breaks the line when it needs to.

src/apps/explorer/pages/Home/index.tsx
```diff
@@ -12,22 +13,27 @@ const Wrapper = styled.div`
...
+      margin-bottom: 1rem;
+    }
+  }
 `
 
 export const Home: React.FC = () => {
   return (
     <Wrapper>
-      <h1>Search Order ID / ETH Address / ENS Address</h1>
+      <h1>Search on Gnosis&nbsp;Protocol&nbsp;Explorer</h1>
```
alongoni, line 36: One more comment here. Can we remove the "&nbsp;" and use a simple space?

src/apps/explorer/components/common/Search/index.tsx
alfetopito: Understood Maria, thanks for the explanation :)

src/apps/explorer/pages/Home/index.tsx
maria-vslvn: We can, but i did it to keep "Gnosis Protocol Explorer " in one line on small mobile screens

src/apps/explorer/components/common/Search/index.tsx
maria-vslvn: thanks!

src/apps/explorer/pages/Home/index.tsx
maria-vslvn: yes, done, thanks!

src/apps/explorer/pages/Home/index.tsx
alongoni: Let's try to decrease the font-size on mobile to fit the text in one line. Then we can remove the `&nbsp;`

elena-zh:
The latest changes LGTM

alongoni (approved):
Looks great!

alfetopito (approved):

PR #881, opened 2021-11-30
(bugfix) Scroll jumping

### LINKED ISSUE #845: [2.4.0] Page is blinking (scroll appears) when order list is updated
It is hard to detect, but when the screen is opened, I can notice that right scrollbar appears when order list is updated each 30 seconds. So it seems like the screen is blinking. 
See the video from 11 sec: https://watch.screencastify.com/v/wlAaVbu2cn7IVgaqL9Ye

alongoni:
# Summary

Closes #845 

# To Test

1. <<Step one>> Open the page `about`

### DIFF
--- a/src/apps/explorer/const.ts
+++ b/src/apps/explorer/const.ts
@@ -2,7 +2,7 @@ import { AnalyticsDimension, Network } from 'types'
 
 /** Explorer app constants */
 export const ORDER_QUERY_INTERVAL = 10000 // in ms
-export const ORDERS_QUERY_INTERVAL = 30000 // in ms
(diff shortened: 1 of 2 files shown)


alongoni:
Well, after analyzing the cause of the problem with @henrypalacios , we found that vertical scrolling appears when we use the TextWithTooltip component. In addition, another strange behavior was identified with the TokenDisplay (imageAddress) which causes a flashing with the data. It is somewhat more complex than I expected. Maybe @matextrem can help here 🙌 

https://www.loom.com/share/4506c2879d5746068bd8e3cdef81c29a

henrypalacios:
@matextrem did you see any of this?

matextrem:
@henrypalacios No, I had no time to take a look at this.

PR #930, opened 2021-12-27
[Explorer] Custom hook for getting explorer data for BlockExplorerLink component when it is not a gpv2 tx

### LINKED ISSUE #857: Transaction viewer page: Identify when tx is/is not a GPv2 transaction (offer link to blockexplorer)
Related to #780

Identify when tx is/is not a GPv2 transaction (offer link to blockexplorer)

matextrem:
# Summary

Closes #857 

Added a custom hook `useTxOrderExplorerLink` that returns the data needed to be used in `BlockExplorerLink` component in case it is not a Gpv2 tx otherwise returns an empty object.

# To Test

Since tx table is not ready yet, there's no UI to reproduce this behaviour. You can test it by importing `useTxOrderExplorerLink` from `hooks/useGetOrders` and use it with a `tx` hash. i.e:  
`const explorerLinkData = useTxOrderExplorerLink('0xd51f28edffcaaa76be4a22f6375ad289272c037f3cc072345676e88d92ced8b5')
`

And then use this data in `BlockExplorerLink` component:

`<BlockExplorerLink {...explorerLinkData} />`


src/hooks/useGetOrders.tsx
```diff
@@ -117,6 +118,42 @@ export function useGetTxOrders(txHash: string): Result {
...
+  const [explorerLink, setExplorerLink] = useState({})
+
+  const fetchTxOrders = useCallback(
+    async (_txHash: string, networkId: Network): Promise<void> => {
+      const { config: networkIdList } = CONFIG.exchangeContractConfig
+      const availableNetworks = new Set(networkIdList.map(({ networkId }) => networkId))
+
+      for (const network of availableNetworks) {
+        try {
+          const ordersFetched = await getTxOrders({ networkId: Number(network), txHash: _txHash })
```
alfetopito, line 133: I believe we'll have to do this check anyway when searching for a given tx.
I'm afraid we'll do several calls to the backend without really needing it.

Do you see what I mean?

matextrem: @alfetopito We have this logic here  https://github.com/gnosis/gp-ui/pull/933/files#diff-92ef89a8db4e59ad32891b00698921ea9e7fac10d9a70c702f0ca0a65322f6c6R195

Once that PR is merged, I'll update this hook in order to get that information directly from the function.

elena-zh:
Hey @matextrem , can we finish this task?

elena-zh:
@matextrem , the only thing I could clarify is that we have already discussed it in the 'parent' issue: https://github.com/gnosis/gp-ui/issues/857#issuecomment-998680635

matextrem:
@elena-zh regarding that comment, the logic is already done. The only thing left is where to show the Explorer link.

alfetopito:
This should be used when searching for a tx by hash.

The steps are basically:
1. Search by a string
2. If it matches txHash regex, checks whether it's a protocol tx
3. If it is, show the tx view
4. Else, check from which network it is (this hook)
5. If it matches any network, return a link to etherscan for given tx
6. Else, search not found.

For example, if I search right now for a non-protocol tx I get a not found https://explorer.cow.fi/search/0x46abe93685cc328f908f44695cd2cc2e65bf75c45b69fc787867468e8735653c
This hook enhances it a little bit by offering the link to the correct block explorer of an existing tx

alfetopito:
Regarding where to display it, I think the appropriate place is in the not found page with a sentence like:

```
This is not a CowProtocol transaction.
See it in the <a>block explorer</a>
```

Feel free to propose a better sentence

elena-zh:
Hey @matextrem , great changes!

However, I'd move the message outside the search block area
ouside

Besides, I'd add an etherscan icon to the link like we do for the all the rest external links in the Explorer
Screenshot_3
and showed 'etherscan' instead of the cropped hash for TxHash from Etherscan/Rinkeby, and 'blocscout' for TxHash in GC.
expl

Also, I noticed that the app navigates always to Etherscan for external transactions: I searched for GC TXHash, and was navigated to the 404 page in Etherscan: 
blocscout

I searched for this TxHash https://blockscout.com/xdai/mainnet/tx/0xb97a644f31364f47efa2a8e3cb98b2c6ed29846aa6a71f514087523b1f2df845
The same is for Rinkeby transactions. 

Could you please fix it?

elena-zh:
Hey @matextrem , looks much-much better!
However, the app does not offer me a link to the Blockscout when I search for an external transaction: 
image

Also, it would be nice to increase margin between lines in the mobile view:
increase margin

matextrem:
@elena-zh I've just tested it and it is working for me. Maybe you have it cached? 

image

Besides, I'm working on your feedback regarding styling.

elena-zh:
@matextrem , hmm, thanks.  Will check. 
I will retest it when you merge you latest changes then.

matextrem:
@elena-zh  done!

elena-zh (approved):
@matextrem , changes LGTM now. 
Thank you!

src/components/orders/OrderNotFound/index.tsx
```diff
@@ -118,6 +127,11 @@ export const OrderAddressNotFound: React.FC = (): JSX.Element => {
           </SearchContent>
         </SearchSection>
       </Content>
+      {showLinkData && (
+        <LinkData>
+          This is not a CowProtocol transaction. See it in the <BlockExplorerLink {...(data as never)} />
```
alfetopito, line 132:
```suggestion
          This is not a CowProtocol transaction. See it on <BlockExplorerLink {...(data as never)} />
```

alfetopito: Also, why casting `data` as `never`?

src/hooks/useGetOrders.tsx
```diff
@@ -149,6 +152,43 @@ export function useGetTxOrders(txHash: string): GetTxOrdersResult {
...
+      for (const network of NETWORK_ID_SEARCH_LIST) {
+        //update provider to find tx in network
+        updateWeb3Provider(web3, network)
+        web3.eth.getTransaction(txHash).then((tx) => {
+          if (tx) {
+            setExplorerLink({
+              type: 'tx',
+              networkId: network,
+              identifier: txHash,
+              showLogo: true,
```
alfetopito, line 177: I'm not convinced on showing Etherscan logo for blockscout, although not strongly against it
Screen Shot 2022-02-23 at 12 46 07

alfetopito (approved):
Minor comments, looking good otherwise

src/components/orders/OrderNotFound/index.tsx
matextrem: @alfetopito I used  `never` in order to keep data as  a generic type in case it is used as a property for sharing extra information to the component. I used `any` at first but eslint recommend me to cast it to `never`

src/hooks/useGetOrders.tsx
elena-zh: > I'm not convinced on showing Etherscan logo for blockscout, although not strongly against it Screen Shot 2022-02-23 at 12 46 07

Hey @alfetopito , then, we need to remove this logo from the footer links, and transaction/user address from the eponymous pages
image

And again, we use Etherscan icons in the Cowswap for GC
image

src/hooks/useGetOrders.tsx
```diff
@@ -149,6 +152,43 @@ export function useGetTxOrders(txHash: string): GetTxOrdersResult {
...
+): ExplorerLinkProps | Record<string, unknown> {
+  const networkId = useNetworkId() || undefined
+  const [explorerLink, setExplorerLink] = useState<ExplorerLinkProps | Record<string, unknown>>({})
+
+  useEffect(() => {
+    if (!networkId) {
+      return
+    }
+
+    if (!orders?.length) {
```
henrypalacios, line 167: Nice solution,

I see that you only need the number of orders here, maybe use something like `isZeroOrders` instead of `orders: Order[]` would win:
1. improve readability with less conditionals added.
```javascript
if (!networkId || !isZeroOrders) return

for (const ...
.
.
.
```
2. Use **primitive** value in the hook.

elena-zh:
LGTM now!

henrypalacios (approved):
:+1: 

@matextrem Just Nitpick: How about not [explicitly using `undefined`](https://github.com/gnosis/gp-ui/pull/930/files#diff-4d74821f906477ec8f10b3221b7f06658bb6f49fa57c08474b58623436d628c9R160) in the state initialization in order to be consistent with the rest of the state definitions.

PR #992, opened 2022-01-17
notifications

### LINKED ISSUE #347: Failed to fetch notification
I got the following message below the order details
Bildschirmfoto 2021-03-30 um 10 31 28

It appears and disappears. After refreshing the website only the error message was shown

maria-vslvn:
# Summary

Closes #347 

In order to reproduce this issue:

1. Create an order
2. While the order is in the 'Open' status, open the order details in the Explorer
3. Switch off the internet connection


elena-zh:
Hi @maria-vslvn !
The notification itself looks good to me in a desktop view. 

Several notes and questions: 
1. Notification text is different than is supposed to be in https://github.com/gnosis/gp-ui/issues/347#issuecomment-941013277 . Besides, there is no 'try again' link that should refresh the page. 
2. I was not able to see error messages on User details page. In the future, we will need to show it on Tx details as well. 
order list

3. I was not able to see this warning (steps to reproduce are here https://github.com/gnosis/gp-ui/issues/347#issuecomment-934840942) 
image

4. The warning is not adjusted to a mobile/tablet views
mobile

5. It would be nice to add margins between text lines in the notification
magrings

elena-zh (changes requested):
Noted above

maria-vslvn:
@alfetopito @ramirotw i had to create to types of error display, but the object i got for errors was like {string: string}. And i just displayed the text. So the only way i thought to separate network error to not network one was with `navigator.onLine`. I'm pretty not sure about the way i separated the errors. So i think i need your help here, or any thoughts, because i'm mostly sure it's wrong:) Also @elena-zh said that the text is different from what was on the mockups, so should i display what comes from erorrs, or change it somehow before display?
@alongoni i added some icons as well, from fontawesome, but they were absent in the modules, so i added, i think we can have something like this for keeping custom icons?

src/components/Notification.tsx
```diff
@@ -0,0 +1,90 @@
+import React, { useState } from 'react'
+import styled from 'styled-components'
+import { transparentize } from 'polished'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faExclamationEllipsis, faExclamationTriangle } from './icons'
+import { BASE_COLOURS } from 'theme'
+
+interface NotificationProps {
+  type: string
```
alfetopito, line 9:
```suggestion
  type: 'warn' | 'error'
```

src/components/Notification.tsx
```diff
@@ -0,0 +1,90 @@
...
+  &:before {
+    left: 50%;
+    transform: rotate(45deg);
+  }
+`
+
+export const Notification: React.FC<NotificationProps> = ({ type, message }: NotificationProps) => {
+  const [isNoteActive, setIsNoteActive] = useState(true)
+
+  const icon = type === 'error' ? faExclamationEllipsis : faExclamationTriangle
```
alfetopito, line 78: I think it'll be cleaner if you add a flag `isError`:

```ts
const isError = type === 'error'
```

src/components/Notification.tsx
```diff
@@ -0,0 +1,90 @@
...
+  return (
+    <NotificationWrap type={type} isActive={isNoteActive}>
+      <FontAwesomeIcon icon={icon} />
+      <span>
+        {message}. Please&nbsp;
+        <a onClick={(): void => window.location.reload()}>{type === 'error' ? 'try again ' : 'refresh '}</a>
+        {type === 'error' ? 'later.' : 'to see new stuff.'}
```
alfetopito, line 85: Not the most professional sentence, hehe. Let's think of something better:

```suggestion
        {type === 'error' ? 'later.' : 'to see get the latest.'}
```

src/components/orders/OrderDetails/index.tsx
```diff
@@ -24,7 +25,7 @@ export type Props = {
   trades: Trade[]
   isOrderLoading: boolean
   areTradesLoading: boolean
-  errors: Record<string, string>
+  errors: any
```
alfetopito, line 28: The best thing here I think is to define the error type and when creating it noting whether it's and error or a warning.

The type would be something like

```ts
export type UiError = {
  message: string
  type: 'warn' | 'error'
}
```

And another type here for aggregating all errors. 
Don't remember why exactly it was created as a Record, so we could change it to an Array, but anyway:

```ts
type Errors = Record<string, UiError>
```

maria-vslvn:
@henrypalacios could you please take a look at this issue? i'm having issues at the point of how should i work with the error new type, how could i get or calculate the type and so on

ramirotw:
@elena-zh please re-test.

- User Details now has the notifications on top
- I added a new ConnectionStatus component which listens to the browser's internet connectivity and displays itself whenever the connection is down. This will give he user more info as to why any request could have failed
- I also shortened the order id in the notification error so it doesn't brake in mobile

ramirotw:
@alfetopito the build fails with:

```
ERROR in Sentry CLI Plugin: Command failed: /home/runner/work/gp-ui/gp-ui/node_modules/@sentry/cli/sentry-cli releases new gp-explorer@v2.6.0
error: API request failed
  caused by: sentry reported an error: You do not have permission to perform this action. (http status: 403)
```

any idea why?

alfetopito:
> @alfetopito the build fails with:
> 
> ```
> ERROR in Sentry CLI Plugin: Command failed: /home/runner/work/gp-ui/gp-ui/node_modules/@sentry/cli/sentry-cli releases new gp-explorer@v2.6.0
> error: API request failed
>   caused by: sentry reported an error: You do not have permission to perform this action. (http status: 403)
> ```
> 
> any idea why?

Yeah, I think so.
We are migrating to our new email accounts, and our old accounts have been removed from Sentry.
But, access tokens were bound to user accounts, so all existing tokens no longer work.

We'll need to update sentry auth tokens before it succeeds.

@W3stside or @anxolin are working on a fix.

elena-zh:
@ramirotw , changes LGTM!
Notifications' texts are different from the ones in the mock-up, but they are clear and seems understandable to a user, frim my perspective. But let wait for @alfetopito tests.

Tiny issue: the banner is blinking each time when the app is trying to update its state: https://watch.screencastify.com/v/87m3yXRRQNZUcNEWsCxx 
Could we do anything with this? 

Also, we have added  a new page to the Explorer: Transactions. So it would be nice to see these banners there as well. 
Let me know please if I need to create a separate issue for this. 
Thanks

ramirotw:
@elena-zh both issues should be fixed

src/components/ConnectionStatus.tsx
```diff
@@ -0,0 +1,22 @@
...
+    window.addEventListener('online', handleConnectionChange)
+    window.addEventListener('offline', handleConnectionChange)
+
+    return (): void => {
+      window.removeEventListener('online', handleConnectionChange)
+      window.removeEventListener('offline', handleConnectionChange)
+    }
+  }, [handleConnectionChange])
+
+  return online ? null : <Notification type="warn" message="Your internet connection is down." appendMessage={false} />
```
alfetopito, line 21: I think this message is maybe assuming too much.

What about using the same message as Cowswap/Uniswap? https://github.com/gnosis/cowswap/blob/ac962fb00d457bc2c4f59432d7d6d7741443dfea/src/components/Header/ChainConnectivityWarning.tsx#L62

```suggestion
  return online ? null : <Notification type="warn" message="You may have lost your network connection." appendMessage={false} />
```

src/components/orders/OrderDetails/index.tsx
```diff
@@ -18,13 +21,15 @@ const TitleUid = styled(RowWithCopyButton)`
   display: flex;
   align-items: center;
 `
+//TODO: use the type
+// type Errors = Record<string, NotificationProps>
```
alfetopito, line 25: Still relevant?

alfetopito (approved):
Minor comments regarding the code.

Regarding the storybook, it's a bit odd. It shows up and quickly disappears

https://user-images.githubusercontent.com/43217/159375686-f612ed6c-7242-4a2c-a746-f9fa7ebb8f31.mov

Not high priority though. If this can be fixed quickly go for it, otherwise I'm fine leaving as is.

ramirotw:
> Regarding the storybook, it's a bit odd. It shows up and quickly disappears

@alfetopito fixed

elena-zh (approved):
Changes LGTM now!

PR #1026, opened 2022-02-02
Remove margins left/right from header

### LINKED ISSUE #1016: Menu button is clued to the network switcher in a mobile view
Related to #1011 , #953

1. Open the Explorer app in a mobile view
2. Verify the Menu button

**AR**: Menu button is clued to the network switcher
clued

**ER**: the button is not clued

henrypalacios:
# Summary

Closes #1016 

## To test
Go to https://pr1026--gpui.review.gnosisdev.com and check the **header** in mobile 320px resolution view

### DIFF
--- a/src/apps/explorer/styled.ts
+++ b/src/apps/explorer/styled.ts
@@ -46,10 +46,6 @@ export const MainWrapper = styled.div`
   ${media.xSmallDown} {
     max-width: 100%;
     flex-grow: 1;
-    header {
-      margin-left: auto;
-      margin-right: auto;
-    }
     footer {
       flex-direction: column;
       flex-wrap: nowrap;
--- a/src/apps/explorer/styled.ts
+++ b/src/apps/explorer/styled.ts
@@ -46,6 +46,19 @@ export const MainWrapper = styled.div`
   ${media.xSmallDown} {
     max-width: 100%;
     flex-grow: 1;
+
+    header {
+      display: flex;
+      flex-wrap: wrap;
+      flex-direction: row;
+      div.network-selector-container {
+        position: relative;
+        padding-top: 1rem;
+        order: 3;
+        justify-content: right;
+        width: 100%;
+      }
+    }
     footer {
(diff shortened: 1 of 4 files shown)


alfetopito:
Doesn't really fix what seems to be the reported issue
Screen Shot 2022-02-02 at 13 52 23

henrypalacios:
@alfetopito It seems that there are very small resolutions :face_in_clouds: 

What do you think if we move the network selector to the next line with small resolutions?
Selection_447

elena-zh:
Hey @henrypalacios , could we do something like this (adjust these properties) in order to resolve this issue:  https://watch.screencastify.com/v/rdNeImPu8KIItBDFGQSH ?

henrypalacios:
Applying @elena-zh 's suggestion and reducing the image by 2% we can get this :point_down: 
Selection_449

Let me know whay you think.

elena-zh (approved):
Looks great to me!

src/apps/explorer/styled.ts
```diff
@@ -46,9 +46,14 @@ export const MainWrapper = styled.div`
...
+
     header {
-      margin-left: auto;
-      margin-right: auto;
+      img {
+        max-width: 98%;
+      }
+      div.network-selector-container {
+        padding-right: 2rem;
+      }
```
alfetopito, line 56: Wouldn't make sense instead to set that property onto the `SelectorContainer` component instead?

alfetopito (commented):
It's working great now!

Just need to figure out the CSS part

alfetopito (approved):
Looks good now 👍 

I've asked @biocom to review it as well BTW

No need to wait though, if there are any comments you can address in a follow up PR

src/apps/explorer/styled.ts
henrypalacios: Yes, I have taken the responsibility for the style to each component.

biocom (approved):

PR #1080, opened 2022-03-29
hotfix - update sentry settings

W3stside:
This should fix sentry settings causing broken builds


W3stside:
as it's all green i will merge and tag. please postmerge review if u feel it's necessary

